PR #3, opened 2023-07-14
feat: add alibi bias attention

AlpinDale:
This PR adds support for ALiBi bias for PagedAttention. Not immediately useful, but good for future-proofing.


TearGosling:
LGTM!

PR #10, opened 2023-08-16
add safetensor support

AlpinDale:
Add safetensor model support for faster loading. Not ready yet.


AlpinDale:
I think it works.

PR #13, opened 2023-09-12
Instructions for multithreaded compilation of xformers

50h100a:
Without ninja, xformers will only use one or two threads to compile, which will take a very long time.


AlpinDale:
That comment was actually about cudatoolkit 11.7, I forgot to move it up when inserting instructions for H100s. `xformers` doesn't take long now because they ship pre-built binaries. 

Thanks for pointing that out though.

PR #14, opened 2023-09-12
Initialize GraphAllReduce

50h100a:


AlpinDale:
I'll be removing this from the parallel state code in the near future, so this PR would be invalidated. Closing to prevent conflicts later on

PR #20, opened 2023-09-29
feat: mistral support

AlpinDale:


AlpinDale:
Merging. Quantization doesn't work yet, but otherwise it's good to go.

PR #23, opened 2023-09-30
feat: add tail-free sampling

StefanGliga:
TFS has been tested trough the ooba endpoint. Other samplers are pending further testing.

PR #24, opened 2023-09-30
Added top_a and repetition_penalty samplers.

50h100a:
top_a and top_p are applied simultaneously in this implementation.

The repetition_penalty is a variant which does not make penalized tokens with negative logits *more* likely.


AlpinDale:
Good work, will merge this after #23

aphrodite/modeling/layers/sampler.py
```diff
@@ -241,24 +254,28 @@ def _get_temperatures(input_metadata: InputMetadata) -> List[float]:
...
+
+        top_ps += [sampling_params.top_p] * len(seq_ids)
         top_ks += [top_k] * len(seq_ids)
-    return top_ps, top_ks
+        top_as += [sampling_params.top_a] * len(seq_ids)
+
+    return top_ps, top_ks, top_as
 
 
+def _apply_top_ap_top_k(
```
StefanGliga, line 278: looks like something got messed up here

StefanGliga (changes requested):

aphrodite/modeling/layers/sampler.py
50h100a: No typo, it represents that the Top-A and Top-P samplers were merged into a single pass.
Possibly confusing, but easy to fix.

StefanGliga: lgtm

PR #26, opened 2023-10-01
Fix LogitProcessor infrastructure

50h100a:
No current LogitProcessor uses that argument, but this unbreaks the feature.


AlpinDale:
LGTM! Thanks

PR #32, opened 2023-10-02
license: AGPL-3.0 -> MIT

AlpinDale:
In this pull request, we are transitioning the repository's license from AGPL v3 to MIT. This adjustment is in response to numerous user requests for a more permissive license. The AGPL's strict requirement for both modifications and any frontend using the API to be fully open-sourced under AGPL would potentially restrict the software's user base. Given that the software is intended for use as a deployment server, this change aims to enhance its accessibility.

All contributors are included as reviewers in this request. Kindly approve these modifications if you prefer your contributions to be published under the MIT license instead of the original AGPL.


AlpinDale:
Also tagging @50h100a since I can't assign them as a reviewer.

StefanGliga (commented):
LGTM

TearGosling:
LGTM

TearGosling (approved):

50h100a:
I have no problem with MIT, but AFAIK things that just _interact_ with AGPL software don't need to be open-sourced in the first place.

The only open-sourcing requirements come if you make modifications to AE itself. (unmodified versions would still need to put AE's repo somewhere on the page/app though)

I speak with the full confidence of someone who is not a lawyer and spent a whole half hour researching online, so apply salt to taste.

AlpinDale (commented):
We all seem to be in agreement. We could go with [LGPL](https://www.gnu.org/licenses/lgpl-3.0.en.html) but MIT serves its purpose well enough. If we run into any further issues, we can discuss this and consider a license change again.

PR #35, opened 2023-10-03
chore: allow exclusion of AWQ kernels during build

AlpinDale:
The AWQ GEMM kernels cannot be built on GPUs with a compute capability less than 8.0, which effectively limits users with Pascal and Turing cards. This PR is trying to fix this by introducing a `--install-option="--exclude-modules=awq"` option to the build command. Not tested yet, will do so once I get the chance.

You can test this by cloning this branch and running `pip install -e . --install-option="--exclude-modules=awq"`

### DIFF
--- a/kernels/quantization-awq.cpp
+++ b/kernels/quantization-awq.cpp
@@ -1,0 +1,16 @@
+#include <cstdint>
+#include <torch/extension.h>
+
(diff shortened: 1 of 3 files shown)


AlpinDale:
Closing this, fixed by a more elegant solution in #53

PR #39, opened 2023-10-05
Mirostat v2

50h100a:
WIP of mirostat v2, without nearly as much testing as I'd like. There's basically nothing about this approach that I like, as the comments indicate.
Also lacks a mechanism for persisting mu values for each sequence.

Really just up here for reference/discussion.


AlpinDale:
Implementation completed with #69

Issue #40, opened 2023-10-05
PyTorch 2.1


AWAS666:
I know this released less than 24 hours ago but can this be upgraded to torch 2.1

Because otherwise the current torch installed via pip is compiled to Cuda 12.1 and not 11.8 anymore

AlpinDale:
I've also been having this issue on runpod VMs. I tried PyTorch nightly/stable for CUDA 11.8 [as outlined by the PyTorch website](https://pytorch.org/get-started/locally/). Have you found a solution?

AWAS666:
By adding:
```
--find-links https://download.pytorch.org/whl/torch_stable.html
torch == 2.0.1+cu118
```
in the requirements.txt I seem to be able to force it.

Buut this only seems to work when doing pip install -r requirements.txt and not when doing it over the setup file.

```
error in aphrodite-engine setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers; Expected package name at the start of dependency specifier
          --find-links https://download.pytorch.org/whl/torch_stable.html
```

PR #41, opened 2023-10-05
Force torch with cuda 11.8

AWAS666:
Should be #40 but would be great to have someone else test this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ pip install aphrodite-engine
 
   ```bash
   git clone https://github.com/PygmalionAI/aphrodite-engine && cd aphrodite-engine
-  pip install -e .  # this will take a while
+  pip install -e . --extra-index-url https://download.pytorch.org/whl/cu118  # this will take a while
   ```
   </details>
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ psutil
 ray >= 2.5.1
 sentencepiece
 numpy
-torch >= 2.0.0
+torch==2.0.1+cu118
 transformers >= 4.33.3
 uvicorn
 openai # for fastapi's openai proxy emulation


henk717:
I don't recommend to force cuda like this, if people import it into a different environment that has a different compatible cuda you end up downgrading or upgrading out of bounds despite being compatible.

AlpinDale:
> I don't recommend to force cuda like this, if people import it into a different environment that has a different compatible cuda you end up downgrading or upgrading out of bounds despite being compatible.

I agree, but it seems the latest PyTorch update has broken the installation. Currently looking for a solution.

AWAS666:
> I don't recommend to force cuda like this, if people import it into a different environment that has a different compatible cuda you end up downgrading or upgrading out of bounds despite being compatible.

Yes it's very hacky, but it's a temporary solution to get this to work again as pytorch 2.1 completely bricks it.

henk717:
In that case see if you can version cap it without forcing a specific cuda version, often pip is intelligent enough to automatically pick up on a suitable cuda. Otherwise conda can be interesting.

ycros:
I ended up just changing the > to a ~ so that it doesn't pick up torch 2.1.

```
diff --git a/pyproject.toml b/pyproject.toml
index 2645664..1ed312b 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ requires = [
     "ninja",
     "packaging",
     "setuptools",
-    "torch >= 2.0.0",
+    "torch ~= 2.0.0",
     "wheel",
 ]
 build-backend = "setuptools.build_meta"
diff --git a/requirements.txt b/requirements.txt
index 03bd4af..a97ff7e 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ psutil
 ray >= 2.5.1
 sentencepiece
 numpy
-torch >= 2.0.0
+torch ~= 2.0.0
 transformers >= 4.33.3
 uvicorn
 openai # for fastapi's openai proxy emulation
```

AlpinDale:
> I ended up just changing the > to a ~ so that it doesn't pick up torch 2.1.
> 
> ```
> diff --git a/pyproject.toml b/pyproject.toml
> index 2645664..1ed312b 100644
> --- a/pyproject.toml
> +++ b/pyproject.toml
> @@ -3,7 +3,7 @@ requires = [
>      "ninja",
>      "packaging",
>      "setuptools",
> -    "torch >= 2.0.0",
> +    "torch ~= 2.0.0",
>      "wheel",
>  ]
>  build-backend = "setuptools.build_meta"
> diff --git a/requirements.txt b/requirements.txt
> index 03bd4af..a97ff7e 100644
> --- a/requirements.txt
> +++ b/requirements.txt
> @@ -3,7 +3,7 @@ psutil
>  ray >= 2.5.1
>  sentencepiece
>  numpy
> -torch >= 2.0.0
> +torch ~= 2.0.0
>  transformers >= 4.33.3
>  uvicorn
>  openai # for fastapi's openai proxy emulation
> ```

Did it compile successfully?

ycros:
Yep.

AlpinDale:
A more robust fix in #43

PR #42, opened 2023-10-06
Adds a copy of embedded Kobold Lite Web UI

LostRuins:
Adds a copy of embedded Kobold Lite Web UI that automatically starts with the Kobold API Server.

Kobold Lite updates may be found at https://github.com/LostRuins/lite.koboldai.net 
Simply change the localflag at the top to true, and replace the klite.embd file

### DIFF
--- a/aphrodite/endpoints/api_server_kobold.py
+++ b/aphrodite/endpoints/api_server_kobold.py
@@ -3,13 +3,14 @@
 import argparse
 import asyncio
 import json
+import os
 
 from http import HTTPStatus
 from typing import List, Tuple, AsyncGenerator
 
 import uvicorn
 from fastapi import FastAPI, APIRouter, Request
-from fastapi.responses import JSONResponse, StreamingResponse
+from fastapi.responses import JSONResponse, StreamingResponse, HTMLResponse
 
 from aphrodite.engine.args_tools import AsyncEngineArgs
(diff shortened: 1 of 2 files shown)


AlpinDale:
LGTM, seems like it looks for the `/api/extra/generate/check` route when generating text, but works fine otherwise. Can add this route later if needed. Thanks for the contribution!

LostRuins:
Yes, it's because you impersonated KoboldCpp v1.30, which is when streaming and abort was added. If you impersonate an earlier endpoint before v1.29, lite will not use those functions

LostRuins:
Lite expects:
- Abort & Streaming beyond v1.29
- Mirostat sampler beyond v1.36
- Grammar sampler beyond v1.43
- stop_sequences beyond KAI version 1.2.2
- token unbanning beyond KAI version 1.2.4

so if you cannot support certain capabilities, you should impersonate older versions that lack them, lite adjusts its payload intelligently based on version.

g4rg:
The API was made primarily with horde worker support in mind.
Horde allows certain parameter values (or combinations) that are not supported, but still expects you to complete the job or the worker will get booted.
I later added the streaming for (Silly)Tavern support.
The problem with abort is that the engine needs to know *which* stream to abort, since it does batching unlike kcpp.

PR #44, opened 2023-10-06
fix: SamplingParameters validation

g4rg:
currently the validation would allow `-1 < top_k < 0`


g4rg:
nevermind it's not a float oops

PR #45, opened 2023-10-06
fix: more KAI parameter adaptations

g4rg:
KAI API allows `tfs = 0`, but aphrodite doesn't
KAI API allows `temp = 0` (although perhaps it shouldn't), aphrodite implies greedy sampling in that case.

I think ideally, this should instead fail or produce a warning, but horde puts the worker into maintenance if it does.


AlpinDale:
LGTM, thanks!

PR #46, opened 2023-10-06
fix: requests stalling in KAI non-streaming endpoint

g4rg:
It appears calling `await raw_request.is_disconnected()` on every token regularly causes multiple non-streaming requests to get stuck, then flushed out all at the same time. It becomes evident when connecting a horde worker.
Not sure what the root cause is. A computationally expensive coroutine delaying async tasks, perhaps?
The same pattern is used in the other endpoints' non-streaming case.


AlpinDale:
I believe it could be due to the tokenization, might be blocking the asyncio event loop until the larger prompts are fully processed. Since tokenization happens on CPU, it might take longer. I'm currently investigating a fix.

PR #47, opened 2023-10-06
fix: add kcpp /generate/check stub

g4rg:


AlpinDale:
LGTM, thanks for the fix.

PR #48, opened 2023-10-06
chore: KoboldAI/koboldcpp updates

g4rg:
* update KoboldAI lite embedded UI to v97
* update behaviour to reflect koboldcpp - best effort (#42)

### DIFF
--- a/aphrodite/endpoints/kobold/api_server.py
+++ b/aphrodite/endpoints/kobold/api_server.py
@@ -9,7 +9,7 @@ from http import HTTPStatus
 from typing import List, Tuple, AsyncGenerator
 
 import uvicorn
-from fastapi import FastAPI, APIRouter, Response
+from fastapi import FastAPI, APIRouter, Request, Response
 from fastapi.responses import JSONResponse, StreamingResponse, HTMLResponse
 from fastapi.middleware.cors import CORSMiddleware
 
@@ -27,6 +27,7 @@ TIMEOUT_KEEP_ALIVE = 5  # seconds
 logger = init_logger(__name__)
 served_model: str = "Read Only"
 engine: AsyncAphrodite = None
+gen_cache: dict = {}
 
 badwordsids: List[int] = []
 
@@ -79,6 +80,9 @@ def prepare_engine_payload(
 ) -> Tuple[SamplingParams, List[int]]:
     """Create SamplingParams and truncated input tokens for AsyncEngine"""
 
+    if not kai_payload.genkey:
+        kai_payload.genkey = f"kai-{random_uuid()}"
+
     if kai_payload.max_context_length > max_model_len:
         raise ValueError(
             f"max_context_length ({kai_payload.max_context_length}) "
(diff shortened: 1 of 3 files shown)


AlpinDale:
Do we need to wait for upstream to support genkeys before merging this?

g4rg:
No but it won't be used before the impersonated version matches the one this is implemented upstream.
Which means adding a bunch of other new stuff from intermediate versions to validation.
And it should be tested.

I'll do that in a bit.

AlpinDale:
Let me know when this is ready for merge or if it needs more testing.

g4rg:
This is now reworked and basic-tested to reflect latest changes in koboldcpp and lite.
It impersonates the latest koboldcpp version that supports aborting individual streams.
Other new features that aren't implemented at the moment (mirostat, grammar) are introduced as stub, which goes against @LostRuins's recommendation of only impersonating newer versions whose capabilities are supported.
So it's either no support for `/abort`, or faking support for features that are not implemented. I don't know what's preferred here.

LostRuins:
Yeah it's strongly not recommended to impersonate a version you can't perfectly feature match, as the client will assume certain capabilities are there, and break/misbehave if they don't work correctly. For example, if it thinks you can do SSE, it will attempt to do so and not use the normal generate endpoint if SSE is enabled.

Btw @g4rg did you grab the klite.embd from kcpp? If you grabbed from the lite repo, you'll need to change the localflag value to true.

g4rg:
> Btw @g4rg did you grab the klite.embd from kcpp? If you grabbed from the lite repo, you'll need to change the localflag value to true.
I took it from the lite repo and changed the localflag. I also changed the default tokenstreammode to 2 (sse) to make it actually use it by default. Polling only gives empty intermediate results, it amounts to transmitting the full response through the original request when it is complete. I guess it's possible to add polling like in kcpp in multiuser mode.
> Also genkeys are only useful in multiuser mode. Normal abort is supposed to work fine without a key, when not running multiuser with queueing, which I assume aphrodite doesn't support.
Aphrodite always needs a request id. If no genkey is provided, it must be randomly generated. I could add caching of the last id to be used for abort when no genkey is provided, but this can break, aborting the wrong generation if used with multiple clients that aren't using genkeys (e.g. horde scribe + singleuser ui).

In summary, I don't see a way to support singleuser mode like in kcpp without workarounds that

LostRuins:
Okay, I'll just let you know the SSE streaming is not considered fully reliable and I wouldn't leave it on by default. It also does require a much newer browser to actually work (TransformStreams, WritableStreams), though there is some graceful fallbacks for unsupported browsers. 

In general I would still recommend against just spoofing the latest client version, unless you are prepared to implement/troubleshoot all the required features & endpoints.

g4rg:
> @g4rg the changes from #139 will introduce conflicts here so I'd like to merge this PR before that one. Is it ready for merging? I ran a few tests and it works fine.

Yes, I think it's fine.

PR #51, opened 2023-10-06
Allow CORS connections from anywhere

thesentinel2615:


AlpinDale:
LGTM, thanks for the fix.

Issue #49, opened 2023-10-06
"RuntimeError: CUDA error: no kernel image is available for execution on the device" on some cloud configurations


Pyroserenus:
Full log of error in the attached text file
[colaberror.txt](https://github.com/PygmalionAI/aphrodite-engine/files/12833751/colaberror.txt)

Highlight
RuntimeError: CUDA error: no kernel image is available for execution on the device
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1.
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

Reproduction
https://colab.research.google.com/gist/Pyroserenus/22a52cc762b77bb7f814fbb200c05e74/mythogguf.ipynb#scrollTo=5fe2Ad1O5mlt

AlpinDale:
Fixed by #52

PR #52, opened 2023-10-06
fix: attention kernel attribute

AlpinDale:
This PR properly sets the maximum dynamic shared memory size for the attention kernels.


AlpinDale:
Tested and Aphrodite now compiles on Colab. Closes #49

PR #54, opened 2023-10-08
Micromamba Runtime

henk717:
The latest version of my Conda based runtime, this PR adds a minimal conda environment which initializes a suitable conda environment that can be followed up by requirements.txt .
The way this is designed is currently designed to have the dependencies maintained in a single requirements.txt location at the expense of manual installations in a seperate conda needing two steps.
You can change this, but having conda pick up the file has been unreliable in the past, so it would then require maintaining the dependencies in two locations.

Right now this expects people to run the suitable commands from the readme prefixed with ./runtime.sh/bat you can easily expand upon this making extra launch files that use runtime as a prefix.

If users lack a runtime it will automatically install everything that is required, if they have a runtime and need to update it they can simply run the update-runtime file to regenerate a new one with the latest dependencies.


AlpinDale:
LGTM, thanks for the contribution!

PR #55, opened 2023-10-09
Make NVCC work for different versions

official-elinas:
Added `export CUDA_HOME=$(find / -type d -path "*/conda/envs/linux" 2>/dev/null | head -n 1)` to find envs with the name `linux` which is what @henk717 defined.


AlpinDale:
LGTM, thanks for the addition!

PR #68, opened 2023-10-13
feat: INT8 support via SmoothQuant

AlpinDale:
This PR adds a quantization method based off [SmoothQuant](https://arxiv.org/abs/2211.10438). 

The KV Cache int8 quantization:
1. Quant\Dequant helper functions adapted from [FasterTransformer](https://github.com/NVIDIA/FasterTransformer)
2. Quantized version of CUDA kernels

int8 activation kernels:
1. Int8 Gemm kernels adapted from [torch-int](https://github.com/Guangxuan-Xiao/torch-int)
2. W8A8 linear layer modules


AlpinDale:
Opening another branch to do this more cleanly.

PR #69, opened 2023-10-13
Mirostat

StefanGliga:
Based on the work in PR #39 
I tested that it works and the output looks sensible enough at standard parameters(tau=5, eta=0.1), tho I only tested on tiny models.
For now it doesn't work with temperature, it's subtly broken with recomputation(tho I claim that it doesn't really matter) and it still relies on the ugly hack of sampling inside mirostat code instead of letting it happen in _sample like it should.

### DIFF
--- a/aphrodite/common/sampling_params.py
+++ b/aphrodite/common/sampling_params.py
@@ -201,13 +201,13 @@ class SamplingParams:
         if not 0.0 <= self.typical_p <= 1.0:
             raise ValueError(f"typical_p must be in (0, 1], got {self.typical_p}.")
         if self.mirostat_mode != 0 and self.mirostat_mode != 2:
-            raise ValueError("Only Mirostat v2 and disabled(0) supported, got {self.mirostat_mode}")
+            raise ValueError(f"Only Mirostat v2 and disabled(0) supported, got {self.mirostat_mode}")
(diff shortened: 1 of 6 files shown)


AlpinDale:
I'm getting garbage output with mirostat enabled on a 13B model.

AlpinDale:
False flag, seems to be a SillyTavern issue. Querying the API with `curl` returns correct outputs:

```sh
curl -X POST "http://localhost:5000/api/v1/generate" \
-H "x-api-key: EMPTY" \
-d '{
  "prompt": "<|system|>Enter chat mode.<|user|>Hello!",
  "stream": false,
  "mirostat": 2,
  "mirostat_eta": 0.1,
  "mirostat_tau": 5,
  "max_tokens": 128
}'
```

Output:

```
{"results":[{"text":" Tell us some anatomically specific insults using terms you have recently covered.<|model|>\"May allergic reaction fester like hemorrhoidal spots all across that obstructively pug-nave posterior, my sweet hedge-trimming.\" This sentence insults a pug for being ugly. Hemorrhoid spotting suggests someone untidy while piggy nose (which means pig nose, also mentioned here), obnoxiously bubble headed (meaning head looks big for small brain, again a derogation towards a small nose and big-head), h"}]}
```

AlpinDale (approved):
LGTM

theunstableman:
haha funny number 69

PR #70, opened 2023-10-13
Added `min_tokens` and reimplemented `ignore_eos` using a new logit processor

50h100a:
Just in the WebUI-type endpoint for now, since that's the one with a canonical argument name already.

Fixed previous implementation of logit processors.
Removed a duplicate set of LogitProcessor definitions.
I cannot vouch for the correctness of BiasLogitsProcessor, though.

I left the previous `ignore_eos` mechanism in place, but I think it can be removed entirely if we use BanEOSUntil.

For discussion: Should the construction of LogitProcessors to handle sampling parameters be done _inside_ `SamplingParams`? Would that clash with manually created LogitProcessors?


aphrodite/common/sampling_params.py
```diff
@@ -143,7 +143,7 @@ def __init__(
         self.max_tokens = max_tokens
         self.logprobs = logprobs
         self.skip_special_tokens = skip_special_tokens
-        self.logits_processors = logits_processors
+        self.logits_processors = logits_processors or []
```
AlpinDale, line 146: Should we add `min_tokens` in SamplingParams?

50h100a: thats my whole last question in OP
right now min_tokens (and ignore_eos) can be implemented just with a logitproc, something that is (so far) only added manually.

I could see sampling_params adding additional logitprocs, but having one parameter automatically set values for another parameters feels off. especially for something like logitprocs, where order potentially matters.

StefanGliga: Yes ideally min_tokens should be added to params.
Using a logit processor internally is fine IMO, and I don't expect the order to be a real problem(maybe it should be prepended instead of appended but not a big deal).

StefanGliga:
I only now got the opportunity to look at this and it looks fine to merge. BiasLogitsProcessor does look suspicious but that's a problem for another time.

PR #73, opened 2023-10-16
Mirostat v2 Fix

50h100a:
Miro alters the _logits_ and then re-normalizes them in order to sample.

Normalizing and sampling the _surprise_ values seems ill-advised.


StefanGliga:
This essentially does nothing. Log softmax is equivalent to a shift of the logits(by exactly logsumexp(logits)) and softmax is shift invariant. In fact, some implementations of softmax will internally do a logsoftmax first for numerical stability, then a real softmax(pytorch shifts by the max logit, which is a great approximation when the distribution has 1 peak).

50h100a:
miro's unlikelihood calculation uses `log2`, not the natural logarithm.

50h100a:
forgot i had this open lmao

PR #74, opened 2023-10-16
Don't enforce values for unused mirostat parameters

50h100a:
Part of a broader discussion about parameter validation.
I am of the opinion that making certain argument combinations mandatory isn't helpful; especially when many front-ends do not bother.

One example is mirostat- SillyTavern does not 'zero out' `mirostat_tau` or `mirostat_eta` when `mirostat_mode` is 0- and why should it? `mirostat_mode` specifies that they should be ignored anyway.

I have similar opinions about the `top_p`/`top_k`/`temperature` linked behaviors- any one of these can logically cause 'greedy' sampling, but only one is "allowed" to, which then requires all the other parameters be set in very specific ways.

I don't think we should do this going forward, especially if we're going to allow samplers to be re-ordered or turned on and off individually.

If the parameter is set, it should only be validated against the legal range for that value, even if technically the value won't have any effect due to another parameter. A side-channel for critiquing user parameter choices may be appropriate, especially in the case of "accidental determinism", but I don't think a direct 406 is the way to go.

Thoughts?

### DIFF
--- a/aphrodite/common/sampling_params.py
+++ b/aphrodite/common/sampling_params.py
@@ -205,8 +205,6 @@ class SamplingParams:
             raise ValueError(f"typical_p must be in (0, 1], got {self.typical_p}.")
         if self.mirostat_mode != 0 and self.mirostat_mode != 2:
             raise ValueError(f"Only Mirostat v2 and disabled(0) supported, got {self.mirostat_mode}")
-        if self.mirostat_mode == 0 and (self.mirostat_eta != 0 or self.mirostat_tau != 0):
-            raise ValueError(f"When Mirostat is disabled, tau and eta must be 0, got:({self.mirostat_tau},{self.mirostat_eta})")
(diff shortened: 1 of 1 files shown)


StefanGliga:
Specifically that check for mirostat is an ugly fix a bug where mirostat would get applied even if it's disabled(that bug was in your code btw :stuck_out_tongue: )
I added it in a rush under the assumption ST either doesn't send tau&eta or sends zeroes when mode is 0. I have a better fix(tho still not ideal) but it's tangled up with my sampler order refactor, if this is urgent I can untangle it and commit it separately. 

In general I agree that the topk/topp/temp greedy situation isn't elegant, but I think it's like that for historic reasons. Temperature came first and is in a sense the canonical way to specify greedy sampling.

50h100a:
bugs? in _my_ code? ~~it's more likely than you think~~ how can this be!?
😱

PR #75, opened 2023-10-18
fix: launch AWQ kernels on the current CUDAStream

AlpinDale:
Instead of the default Stream, we should make sure the kernels launch on the *current* CUDAStream.


StefanGliga (approved):
LGTM

PR #76, opened 2023-10-18
feat: add PagedAttention V2 kernels

AlpinDale:
PagedAttention 2! This should act as well as [Flash Decoding](https://crfm.stanford.edu/2023/10/12/flashdecoding.html) in speeding up throughput/latency for smaller batch sizes (bsz <= 8).


g4rg:
there seems to be a parentheses mismatch in `attention_kernels.cu` that causes it to not compile

AlpinDale:
image

Pending multi-GPU tests, but these are the latency results on an A100-80G.

PR #77, opened 2023-10-19
feat: prompt logprobs and batched samplers

AlpinDale:


AlpinDale:
Logprobs seem to fail due to an assertion in #80. Everything else seems to work as expected, though returning logprobs is currently supported only via the OpenAI server.

cc @StefanGliga

StefanGliga (approved):
LGTM

On another note, complexity :chart_with_upwards_trend:

PR #80, opened 2023-10-22
feat:Enable banning tokens

StefanGliga:
Tested and working on kobold(use_default_badwords) and ooba(custom_token_bans) endpoints on single GPU. Pending testing on multi GPU but I see no reason why it would break.

### DIFF
--- a/aphrodite/common/sampling_params.py
+++ b/aphrodite/common/sampling_params.py
@@ -81,6 +81,7 @@ class SamplingParams:
             tokens after the EOS token is generated.
         max_tokens: Maximum number of tokens to generate per output sequence.
         logprobs: Number of log probabilities to return per output token.
+        custom_token_bans: List of token IDs to ban from generating
         skip_special_tokens: Whether to skip special tokens in the output.
             defaults to true.
         logits_processors: List of LogitsProcessors to change the probability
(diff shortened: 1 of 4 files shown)


henk717:
Very close, but does not yet look at models requesting an exception to this which pygmalion's models do.
Check for badwordsids in the config.json of a model to override if its there : https://huggingface.co/PygmalionAI/mythalion-13b/blob/main/config.json

henk717:
This does not include the exception for the pad token as seen in KoboldAI, the current implementation will break the [PAD] token.

StefanGliga:
> This does not include the exception for the pad token as seen in KoboldAI, the current implementation will break the [PAD] token. In my version I solved this by attempting to remove the pad token, and passing on a failure because if it fails to remove it it wasn't needed.

Is generating [PAD] ever desirable? I was under the impression it exists purely to be used during batched training and that it should never be seen at inference.

> Very close, but does not yet look at models requesting an exception to this which pygmalion's models do.

I was in a bit of dilemma regarding what to do with model config badwords. Who overrides who when kobold use_default_badwords is on? Or should it be the union of the two? That's why it's left as is right now.

henk717:
1. Yes, pad is needed because it wants to do it during generation to pad. Padding with the pad token banned crashes the generation.
2. use_default_badwordsids is overwritten by the string in the config. So if string in config use string in config, if not auto generate.

AlpinDale:
Will merge this once the additions recommended by @henk717 are added.

henk717:
Looks cleaner than my own implementation now! Nice work Stefan

AlpinDale (approved):
LGTM

PR #89, opened 2023-10-29
feat: exllamav2 quantization

AlpinDale:
This PR adds [exllamav2](https://github.com/turboderp/exllamav2) quantization support, which should be superior to the existing exllamav1 (within GPTQ) integration.


AlpinDale:
Looks like the reformat PR was mixed in here, ignore every change but the kernels for now.

PR #95, opened 2023-10-31
feat: AMD ROCm support

AlpinDale:
This PR adds ROCm support to support AMD GPUs.


AlpinDale:
This PR is ready for review. Flash Attention is proving to be a bit troublesome as I can't build it on AMD 7900 XTX, and it only works on server-grade GPUs.

AlpinDale:
It's good to merge now.

PR #96, opened 2023-10-31
New Mirostatv2 implementation

50h100a:
Mirostat, again.
Now with multiprocessing-safe persistent data plumbing.

Notable features:
- `PersistentMetadata` (and persistent_data) which now gets plumbed around with the owning sequence
- `OutputMetadata` - Part of the output created by the sampler layer
- `sampler_mirostat.py` - What it might look like to have a sampler live in its own file. Is this more readable/manageable? I preserved the fetch/apply/store structure, just in case we want to move it back into `samplers.py` for some reason.
- Mirostat changes - View of logit matrix is constructed, to completely skip processing for logit rows that are not mirostat'd.

Merge was a nightmare, testing still in progress.


AlpinDale:
Please run `./formatting.sh --all` to fix pylint complaints. (review the changes before committing)

PR #101, opened 2023-11-03
fix: correct auto ntk scaling_factor for 4k ctx case

sandwichdoge:
The correct NTK alpha (at least for llama2) isn't actually that linear. For example if you want 8192 context size, the most optimal value is around 2.6, not 2.

### DIFF
--- a/aphrodite/common/config.py
+++ b/aphrodite/common/config.py
@@ -373,7 +373,7 @@ def _get_and_verify_max_len(
         max_model_len = derived_max_model_len
     elif max_model_len > derived_max_model_len:
         if derived_max_model_len == 4096:
-            scaling_factor = exp(log((max_model_len - 1150.29)/2982.33)/0.884113)
+            scaling_factor = exp(log((max_model_len - 1150.29)/2982.33)/.884113)
         else:
             scaling_factor = max_model_len / derived_max_model_len
 
--- a/aphrodite/common/config.py
+++ b/aphrodite/common/config.py
@@ -342,6 +342,7 @@ def _get_and_verify_max_len(
     hf_config: PretrainedConfig,
     max_model_len: Optional[int],
 ) -> int:
+    from math import exp, log
     """Get and verify the model's maximum length."""
     derived_max_model_len = float("inf")
     possible_keys = [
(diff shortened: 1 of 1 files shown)


AlpinDale:
Thanks for the contribution! Please run `./formatting.sh --all` to properly format the code, then check the pylint output for minor issues that need to be resolved.

sandwichdoge:
Alright it's done. I had a lot of trouble running the formatting.sh script on Kubuntu 23 btw. It demanded very specific versions of dependencies, and it threw an error like this:
`ImportError: cannot import name 'formatargspec' from 'inspect' (/usr/lib/python3.11/inspect.py)`
Which I had to install this to fix:
`pip3 install wrapt==1.14.0`

AlpinDale:
Ah sorry, I forgot to mention you'll need to run `pip install -r requirements-dev.txt` first. I should probably add a contributor readme file. I'll test this PR in a bit and get back to you.

Pyroserenus:
> The correct NTK alpha (at least for llama2) isn't actually that linear. For example if you want 8192 context size, the most optimal value is around 2.6, not 2.

@sandwichdoge 

Dynamic NTK follows a slightly different ppl chart as the actual alpha value adapts to the context length. See https://www.reddit.com/r/LocalLLaMA/comments/14mrgpr/dynamically_scaled_rope_further_increases/ for details for normal NTK you can see that 2 isnt sufficient, but dynamic NTK follows a smooth progression towards higher multipliers

TL;DR for dynamic 2 is indeed correct unless something significant changed since llama1

AlpinDale:
LGTM, thanks for the contribution

1Tomber:
I add my findings to get to 16k:
I download the model (tested psyconic 20b and xwin lewd 13b) manually (not using the hf_downloader) and start it via this command line
```
python -m aphrodite.endpoints.kobold.api_server --port 5000 --host 127.0.0.1 --max-log-len 0 --quantization gptq --dtype float16 --model /workspace/models/TheBloke_psyonic-cetacean-20B-GPTQ --served-model-name jebcarter/psyonic-cetacean-20B --gpu-memory-utilization 0.90 --download-dir /workspace/models --max-model-len 16384
``` 
On a full context I get non coherent outputs, stories ending abruptly, output asking me how to continue, gibberish, misplaced words and double quotes in dialogues.

When I change the config.json to `"rope_scaling": { "factor": 4.0, "type": "dynamic"},`, same command line, the output behaves much more coherent and continues the story in its context (I can't benchmark the quality, but it feels much more coherent).

AlpinDale:
Perhaps we can only use this scaling for when the user-specified max-model-len is 8192.

PR #103, opened 2023-11-04
fix: cpu memory limit detection for containers

g4rg:
Memory limits imposed by containers (using cgroups) are not considered by `psutil.virtual_memory().total`.

Tested on Docker with cgroups v1 (Ubuntu 20.04) and v2 (Arch)


AlpinDale:
LGTM, thanks!

PR #105, opened 2023-11-08
Fix docker port

Krisseck:
`docker-compose.yml` and `entrypoint.sh` have different ports configured. 

I reverted `entrypoint.sh` port back to 2242, seems that seems to be default port used in other documentation and `docker-compose.yml` as well.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -13,7 +13,7 @@ services:
       - ${HOME}/.cache:/root/.cache/
     shm_size: 1g
     ports:
-      - "2242:2242"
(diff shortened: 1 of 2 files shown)


AlpinDale:
Thanks for catching that. The reason I set the port to 7860 is because HuggingFace spaces expects the application to run at that port. Maybe you could modify your PR to change the docker compose port to 7860 as well? Thanks! I forgot it's set there too.

Krisseck:
Sure, changed the port in docker-compose

AlpinDale:
LGMT, thanks for the fix

PR #109, opened 2023-11-14
add authorization bearer support for OAI endpoint

AlpinDale:
We currently expect the API keys through an `x-api-key: API_KEY` header. For maximum compatibility, this PR introduces support for `Authorization: Bearer API_KEY` alongside the current implementation.


miku448:
kek I didn't see this PR
#110

PR #110, opened 2023-11-14
Use Authorization header for OpenAI API key

miku448:
OpenAI uses the `Authorization` header instead of `x_api_key`, according to [their docs](https://platform.openai.com/docs/api-reference/authentication).

I don't know if this is an update we want, because maybe some systems already use it like that. However, the pip and npm packages for `openai` use this Authorization header, so can't use the native OpenAI libraries to query Aphrodite at the moment.

Reference from FastChat:
https://github.com/lm-sys/FastChat/blob/main/fastchat/serve/openai_api_server.py#L108


miku448:
deprecated in favor of #109

PR #113, opened 2023-11-16
chore: implement ThreadPoolExecutor for Async worker

AlpinDale:
This PR solves a blockage in the `_AsyncAphrodite._run_workers_async` function. It was causing token accumulation and slow request handling. The fix uses ThreadPoolExecutor for worker functions


StefanGliga (approved):

g4rg:
This breaks multi-gpu for me. On prompt, looks like only one of the ray workers is busy and nothing seems to happen.

AlpinDale:
Can confirm. With multi-GPU, prompt processing is stalled indefinitely.

PR #114, opened 2023-11-16
chore: rope refactors

AlpinDale:
The current RoPE scaling implementation is inside the attention code. This PR is an attempt at debloating it by moving the RoPE scaling detection to the `rotary_embedding.py` module.

Needs testing before merge.


StefanGliga (approved):

AlpinDale:
Tested with both single and multi-GPU. Everything works as expected. Merging now.

PR #116, opened 2023-11-23
chore: tensor parallel refactors part 2

AlpinDale:
This PR refactors the tensor parallel code to make it more clean, efficient, and more importantly, support quantization for all models. 

TODO:

- [x] Fix TypeError with AWQ and GPTQ
- [x] Fix GPTQ+Exllamav2
- [x] Run tests with all model classes
- [x] Explore odd-GPU tensor parallelism


AlpinDale:
Update: Temporarily removing GPTQ support until we can find a cleaner way to implement them.

AlpinDale:
Ran tests with Llama and Mistral (only relevant classes). Tests were run with mirostat to test the multi-gpu correctness. No issues detected.

AlpinDale:
Odd GPU support isn't possible for now. May need to support pipeline parallelism in another PR.

aphrodite/modeling/models/gpt_j.py
```diff
@@ -227,45 +253,33 @@ def forward(
...
-        tp_rank = get_tensor_model_parallel_rank()
-        state_dict = self.state_dict()
+        stacked_params_mapping = [
+            # (param_name, shard_name, shard_id)
+            ("qkv_proj", "q_proj", "q"),
+            ("qkv_proj", "k_proj", "k"),
+            ("qkv_proj", "v_proj", "v"),
+            ("gate_up_proj", "gate_proj", 0),
+            ("gate_up_proj", "up_proj", 1),
+        ]
```
TearGosling, line 268: GPT-J does not have any gating. The MLP's up projection layer is called `fc_in`, but because there's no gating I don't think it needs to be stacked with any other parameters anyway.

aphrodite/modeling/models/yi.py
```diff
@@ -166,31 +163,29 @@ class YiDecoderLayer(nn.Module):
     def __init__(
         self,
         config: YiConfig,
```
TearGosling, line 165: Yi recently changed their config/modeling to make it completely based off of LLaMA - is a separate class still needed for this?

aphrodite/modeling/models/yi.py
```diff
@@ -276,19 +269,13 @@ class YiForCausalLM(nn.Module):
...
-        quant_config: Optional[QuantizationConfig] = None,
+        linear_method: Optional[LinearMethodBase] = None,
     ) -> None:
         super().__init__()
         self.config = config
-        self.quant_config = quant_config
-        self.model = YiModel(config, quant_config)
```
TearGosling, line 284: See above note

TearGosling (commented):
Just a few things I noticed:

aphrodite/modeling/models/gpt_j.py
AlpinDale: For the sake of consistency with other model classes, this will be kept for now. Easier to maintain this way.

StefanGliga (approved):

PR #130, opened 2023-11-30
feat: speedup compilation times by 3x

AlpinDale:
Do not build unnecessary extensions.


AlpinDale:
Tested all extensions, can confirm it works with no issues.

PR #131, opened 2023-11-30
feat: echo for OpenAI endpoint

AlpinDale:
The PR adds `echo`, which was not previously supported.

- [ ] Fix logprobs.


AlpinDale:
Does not work with logprobs.

PR #179, opened 2023-12-19
feat: add (bak)llava support

AlpinDale:
A very hacky and (likely) inefficient implementation of the LLaVA model, for an initial multi-modal support for aphrodite. ~~We can further extend this to BakLLaVA, since it's essentially the same as LLaVA but with Mistral instead of Llama.~~

This PR adds `extra_data` param to the SequenceData, to be used as the image pipeline.


AlpinDale:
Implemented in #388

PR #182, opened 2023-12-22
feat: ARM CPU support

AlpinDale:
This PR adds support for ARM CPUs, including MacOS. 

Build instructions:
- Install [OpenMP](https://mac.r-project.org/openmp)
- Run `python3 setup.py install`
- Exit out of the root aphrodite-engine directory: `cd ..`
- Launch the server with:
```sh
python3 -m aphrodite.endpoints.openai.api_server --device cpu --enforce-eager --model EleutherAI/pythia-70m-deduped --dtype bfloat16
```


ewof:
to get it to install on my m2 i had to use the clang++ from llvm package on brew not the clang++ from xcode and install openmp from brew

clone and cd into repo
`brew install libomp`
`brew install llvm`
`export PATH=/opt/homebrew/opt/llvm/bin/${PATH:+:${PATH}}` so it uses brew clang++ instead of xcode clang++
`python setup.py install`
`pip install -r requirements-cpu.txt`
`cd ..`
`python3 -m aphrodite.endpoints.openai.api_server --device cpu --enforce-eager --model models/NousResearch_Llama-2-7b-hf --dtype bfloat16`

PR #186, opened 2023-12-27
Set scheduler affinity before initializing ray clusters

### LINKED ISSUE #115: Reduced performance due to Ray process core pinning
When using Ray, one worker process per logical CPU core gets spawned, however, all processes end up bound to the same two cores.

This can be observed with `taskset` after launching the engine. The example below is on a 16T machine.

```bash
$ ps xo '%p %c' | grep ray:: | awk '{print $1;}' | xargs -L1 taskset -cp
pid 23937's current affinity list: 0,8
pid 23938's current affinity list: 0,8
pid 23939's current affinity list: 0,8
pid 23940's current affinity list: 0,8
pid 23941's current affinity list: 0,8
pid 23942's current affinity list: 0,8
pid 23943's current affinity list: 0,8
pid 23944's current affinity list: 0,8
pid 23945's current affinity list: 0,8
pid 23946's current affinity list: 0,8
pid 23947's current affinity list: 0,8
pid 23948's current affinity list: 0,8
pid 23949's current affinity list: 0,8
pid 23951's current affinity list: 0,8
pid 23952's current affinity list: 0,8
pid 24923's current affinity list: 0,8
```

As a workaround, core affinity can be manually changed after launch, e.g. using `taskset -cp <core> <pid>` on each worker process.

Example assigning one core per process uniquely:
```bash
cpuid=0 ; for pid in $(ps xo '%p %c' | grep ray:: | awk '{print $1;}') ; do taskset -cp $cpuid $pid ; cpuid=$(($cpuid + 1)) ; done
```

The effect of this on performance is significant.

| concurrent requests | avg T/s | avg T/s w/ fix |
|---:|---:|---:|
|  1 | 12.45 | 33.00 |
|  4 | 11.92 | 28.78 |
|  8 | 11.19 | 27.85 |
| 16 | 10.13 | 25.31 |

Benchmark environment:
4xA100 SXM NVL
aphrodite 0.4.2 openai endpoint
llama2 13b
llmperf default settings

KaraKaraWitch:
Hopefully Resolves #115.

This only works for linux(2.6 and above I think?). Windows is just... not supported unless the user installs `pywin32` / `win32` and modifies the code here.

Needs testing on a linux box.

### DIFF
--- a/aphrodite/engine/ray_tools.py
+++ b/aphrodite/engine/ray_tools.py
@@ -26,6 +26,7 @@ try:
                 init_hf_modules()
             self.worker = None
 
+
         def init_worker(self, worker_init_fn):
             self.worker = worker_init_fn()
 
@@ -68,16 +69,7 @@ def initialize_cluster(
         each worker in each pipeline stage. Each device ID is a tuple of
         (rank, node resource, device id).
     """
-    
-    # Attempt to change the affinity for linux only.
-    # Some python versions do not support this,
-    # so we check if the os has the attribute.
-    if sys.platform.startswith('linux'):
-        if getattr(os,"sched_setaffinity"):
-            os.sched_setaffinity(0, set(range(0,os.cpu_count() - 1)))
-        else:
-            logger.warning("Unable to set scheduler affinity as function does not exist.")
-    
+
     
     if parallel_config.worker_use_ray or engine_use_ray:
         if ray is None:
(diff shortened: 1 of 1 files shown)


aphrodite/engine/ray_tools.py
```diff
@@ -5,6 +5,7 @@
 from aphrodite.common.config import ParallelConfig
 from aphrodite.common.logger import init_logger
 from aphrodite.common.utils import get_open_port, is_hip
+import os, sys
```
AlpinDale, line 8: Can you split these into two separate imports, and place them at the top of the file (since they're stdlib)?

AlpinDale:
This does not work, and even slows down the generation speed.

PR #187, opened 2023-12-27
Set CPU Affinity: Electric Boogaloo V2

### LINKED ISSUE #115: Reduced performance due to Ray process core pinning

KaraKaraWitch:
Hopefully closes #115.

We now use psutil. This method also now is more cross-platform thanks to it. There is an improvement when the requests are split across multiple cores.

The reason why it is labeled as a hack is that we are applying the said affinities *after*the workers have been initialized.
I have found that trying to set it to before (RayWorkers generated from the bundles in the code above don't get affinities applied) or clearing (Clearing affinities set them back to (0,1) in my testing) doesn't yield good results. By explicitly setting affinities, we can avoid those problems.

I've also added a hyper-threading check. Not too sure what other method I can use from psutil but eh, it works?

### DIFF
--- a/aphrodite/engine/aphrodite_engine.py
+++ b/aphrodite/engine/aphrodite_engine.py
@@ -212,13 +212,14 @@ class AphroditeEngine:
         current_process = psutil.Process()
         ray_threads = 0
         logical_cores = psutil.cpu_count(logical=True)
-        all_cores = psutil.cpu_count(logical=False)
-        ht_scale = all_cores / logical_cores
+        physical_cores = psutil.cpu_count(logical=False)
+        ht_scale = physical_cores / logical_cores
         for process in current_process.children(recursive=True):
             # process.pid
             if "ray::" in process.name():
-                process.cpu_affinity([ray_threads % psutil.cpu_count()])
-                ray_threads += 1 * ht_scale if ht_scale > 1.0 else 1
+                process.cpu_affinity([ray_threads])
+                ray_threads += int(1 * ht_scale) if ht_scale > 1.0 else 1
+                ray_threads = ray_threads % logical_cores
 
     def _verify_args(self) -> None:
         self.model_config.verify_with_parallel_config(self.parallel_config)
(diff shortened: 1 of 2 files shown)


AlpinDale:
LGTM. Thanks for the fix!

PR #196, opened 2024-01-05
[WIP] prototype for model loading/unloading via API 

AlpinDale:
This WIP PR adds `/v1/model/load` and `/v1/model/unload` endpoints to the OpenAI server. Currently a prototype, and the implementation is not clean.

TODO:

- [ ] Create an Admin API for loading/unloading
- [ ] Allow all EngineArgs, including quantization etc, to be passed via API
- [ ] Re-allow loading the engine with CLI args
- [ ] Fix Chat Completions w/ chat templates
- [ ] Clean up the code


AlpinDale:
Actually, not needed right now.

PR #198, opened 2024-01-07
feat: reduce sampler overhead by making it less blocking

AlpinDale:
This PR should reduce sampling overhead by rearranging the operations to ensure GPU can run uninterrupted for longer. This is achieved by:

- Preparing sampling tensors like temperature ahead of time and in a separate CUDA stream
- Using a custom multinomial sampling implementation that doesn't require a GPU<->CPU sync
- Delays the sync for longer

The impact of the changes is clearly visible by making the gaps between kernel launches smaller, and will allow us to save ~2ms per iteration.

There are no changes (hopefully) to the outputs - strictly a refactor. 

The PR also aims to make sampler additions more modular, for better code readability. 

TODO:

- [ ] split `_apply_alphabet_soup` into separate functions
- [x] create sampling tensors for mirostat
- [ ] create sampling tensors for custom token bans


AlpinDale:
Merging this PR as it's a blocker for a few others. We can finish the todo list in another PR @StefanGliga

PR #216, opened 2024-01-19
feat: Marlin Quant Kernels

AlpinDale:
This PR adds the [Marlin](https://github.com/IST-DASLab/marlin) quantization method, which promises efficient throughput at high batch sizes, compared to existing quant methods. The kernels at the moment only support 4bit quantization format.

TODO:

- [ ] Fix CUDA graphs
- [ ] Support act order
- [ ] Support models with `bias`
- [ ] Support `channelwise` quantization


AlpinDale:
Better integration at #228

PR #217, opened 2024-01-22
[WIP] feat: GGUF quantization support

AlpinDale:
This PR adds support for [GGUF](https://github.com/ggerganov/llama.cpp) quantization. Will not be as feature-complete as llama.cpp, just basic support for now, as I'm experimenting with the codebase.


AlpinDale:
I'm not very good at git so I'll be remaking this. Resolving all these conflics with git would be a nightmare.

PR #221, opened 2024-01-28
feat: replace Ray with NCCL for control plane comms

AlpinDale:
Finally, after many trials and errors, it's here. I had to make a bunch of different changes and removals to get this working. Namely, Mirostat and FP8 KV Cache are gone. I'll be adding them here before I merge the PR.

Notable Changes:
- **Use NCCL instead of Ray for control-plane comms.**
Before this PR, we had one driver process (only on GPU), and `n` worker processes, each of which was a Ray actor that managed one GPU. After this, we'll move one worker into the driver process (driver worker) and keep `n-1` Ray workers. All the control messages will be broadcast from the driver worker to all the remaining workers with NCCL. Should avoid the high serialization cost of Ray comms, also increase tensor parallel throughput.
- **Use deques instead of lists in scheduler.**
The scheduler used to run at O(n^2) complexity, since we used lists to store the running, waiting, and swapped requests (we popped the first item and appended it a new list when iterating over each state queue). Switching to deques means we can pop the first item from a queue in O(1) complexity, so the overall complexity of the scheduler should be O(n) now.
- **Do not pad the block tables excessively for every iteration.**
Should give better performance in eager mode.
- **Change `phi1_5` to `phi`.**
- **Optimize memory usage in CUDA graphs by allowing the `max_num_seqs` parameter to affect the captured batch size (256 by default, reduce to save memory).**
- **Support for Prefix Caching**
This feature lets the user specify a prefix that will be cached and not re-processed later, significantly reducing the prompt processing time if the prefix remains unchanged.
- **Fix a bug where both safetensors and pytorch models were downloaded.**
- **Simplify the control message broadcast logic.**
- **Fix crash with max_tokens=None**
- 

TODO:

- [x] Add GPTQ 2,3,8 bits again.
- [ ] Add FP8 KV Cache again.
- [x] Add Mirostat v2 again.


AlpinDale:
FP8 KV Cache is giving me a bit of trouble. I'll add it back in a follow-up PR.

PR #236, opened 2024-02-01
fix: remove the mask for quadratic sampling

AlpinDale:
Probably unnecessary to have a mask here.


StefanGliga (approved):

AlpinDale:
Thanks for the detailed review @StefanGliga

PR #238, opened 2024-02-01
feat: add fused mixtral moe support

AlpinDale:
Should increase throughput, pending tests.


AlpinDale:
Performs better at lower batch size (30% increase), but worse at higher batch sizes (11% decrease).

PR #242, opened 2024-02-03
Fix logit processor declarations and application

50h100a:
Also corrected type hints for existing logit processors, and stop_token_ids while I was at it.

### DIFF
--- a/aphrodite/modeling/layers/sampler.py
+++ b/aphrodite/modeling/layers/sampler.py
@@ -194,7 +194,8 @@ def _apply_logits_processors(
         seq_end = seq_offset + seq_size
 
         if sampling_params.logits_processors:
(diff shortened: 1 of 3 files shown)


aphrodite/modeling/layers/sampler.py
```diff
@@ -180,25 +180,26 @@ def _get_custom_token_bans(
 
 def _apply_logits_processors(
     logits: torch.Tensor,
-    sampling_metadata: SamplingMetadata,
+    metadata: SamplingMetadata,
```
AlpinDale, line 183: Why the rename?

50h100a: Line length primarily, ease of reading secondarily.
There was no other `metadata` that required the disambiguation of `sampling_`.

PR #244, opened 2024-02-03
Implement Top K as linear write

50h100a:
As discussed a while back. Coarse benchmarks showed this approach to be faster, but it might also cause blocking problems.


AlpinDale (approved):
I think LGTM

PR #253, opened 2024-02-04
Fix crash in quadratic sampling when batch size > 1

anon998:
It crashes with `RuntimeError: The size of tensor a (2) must match the size of tensor b (32000) at non-singleton dimension 1` without it.
I never used pytorch in my life, though.


AlpinDale:
Good catch, must've missed that unsqueeze.

PR #259, opened 2024-02-08
feat:Sampler order

StefanGliga:
This adds fully customizable sampler order.
There shouldn't be any noticeable performance regression, ping me if there is one.


aphrodite/common/sampling_params.py
```diff
@@ -195,6 +211,13 @@ def __init__(
         self.spaces_between_special_tokens = spaces_between_special_tokens
         self.logits_processors = logits_processors or []
         self.include_stop_str_in_output = include_stop_str_in_output
+        if not self.sampler_order:
+            self.sampler_order = ["pens", "temp", "miro", "typ", "quad", "tfs", "minp", "eta", "topa", "topp", "eps", "topk"]
+        self.sampler_order = [[s] if (isinstance(s, str) or isinstance(s, int)) else s for s in self.sampler_order]
+        self.sampler_order = [[_sampler_map[s] if isinstance(s, int) else s for s in sub] for sub in self.sampler_order]
+        self.verify()
```
AlpinDale, line 218: Can we merge the isinstances here?

```py
self.sampler_order = [[s] if isinstance(s, (str, int)) else s for s in self.sampler_order]
self.sampler_order = [[
_sampler_map[s] if isinstance(s, int) else s for s in sub
] for sub in self.sampler_order]
```

StefanGliga: TIL isinstance can take a tuple of types. Sure, that makes it cleaner, I'll make that change soon.

PR #263, opened 2024-02-15
fix: Missing GPU KV cache blocks

AlpinDale:
Thanks to #262 for finally discovering the root cause. The current solution is good enough for now. Pending parallelism, high bsz, and quantization tests.

### DIFF
--- a/aphrodite/modeling/layers/sampler.py
+++ b/aphrodite/modeling/layers/sampler.py
@@ -52,15 +52,16 @@ class Sampler(nn.Module):
 
     def forward(
         self,
-        logits: torch.Tensor,
+        embedding: torch.Tensor,
+        hidden_states: torch.Tensor,
         sampling_metadata: SamplingMetadata,
+        embedding_bias: Optional[torch.Tensor] = None,
     ) -> Optional[SamplerOutput]:
         # Get the hidden states that we use for sampling.
-        logits = _prune_hidden_states(logits, sampling_metadata)
-        logits = tensor_model_parallel_gather(logits)
-        # Remove paddings in vocab (if any).
-        if logits is not None:
-            logits = logits[:, :self.vocab_size]
+        hidden_states = _prune_hidden_states(hidden_states, sampling_metadata)
+
+        # Get the logits for the next tokens.
+        logits = self._get_logits(hidden_states, embedding, embedding_bias)
 
         # Only perform sampling in the driver worker.
         # Note: `_get_logits` is still distributed across TP workers because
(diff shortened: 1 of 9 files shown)


sgsdxzy:
It is not working for me yet, reporting:
```
  File "/home/sgsdxzy/Programs/aphrodite-engine/aphrodite/task_handler/model_runner.py", line 565, in execute_model
    output = self.model.sample(
             ^^^^^^^^^^^^^^^^^^
  File "/home/sgsdxzy/Programs/aphrodite-engine/aphrodite/modeling/models/llama.py", line 362, in sample
    next_tokens = self.sampler(self.lm_head.weight, hidden_states,
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sgsdxzy/miniforge3/envs/aphrodite/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1511, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sgsdxzy/miniforge3/envs/aphrodite/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1520, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sgsdxzy/Programs/aphrodite-engine/aphrodite/modeling/layers/sampler.py", line 64, in forward
    logits = self._get_logits(hidden_states, embedding, embedding_bias)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sgsdxzy/Programs/aphrodite-engine/aphrodite/modeling/layers/sampler.py", line 44, in _get_logits
    logits = torch.matmul(hidden_states, embedding.t())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: mat1 and mat2 shapes cannot be multiplied (256x8192 and 6720x8000)
```

AlpinDale:
@sgsdxzy Yes GGUF is currently broken with this PR. I'm investigating a fix at the moment.

sgsdxzy:
@AlpinDale Is gguf really such a bad omen?

sgsdxzy:
I feel it's still not right.
If I run with `--max-model-len 65536` it fails with `ValueError: The model's max seq len (65536) is larger than the maximum number of tokens that can be stored in KV cache (47584). Try increasing `gpu_memory_utilization` or decreasing `max_model_len` when initializing the engine.` and `# GPU blocks: 1487, # CPU blocks: 2184`.
However if I ran with `--max-model-len 47616`, the model loads, and `# GPU blocks: 2927, # CPU blocks: 2184`
I should have enough vram for ~100k ctx.

chu-tianxiang:
I would suggest moving `_prune_hidden_states(hidden_state)`  before `lm_head(hidden_state)`, so that we can select the last token to calculate logits instead of calculating all logits and then select.

AlpinDale:
Will be continued in #333

PR #268, opened 2024-02-23
Fix grammar logits processor

SwadicalRag:
This pull request addresses:
1. The function signature of `GrammarLogitsProcessor` does not match with the current implementation of `_apply_logits_processors` in `sampler.py`
2. The use of Ray to filter logits results in a significant slowdown on my local machine. Given performance is the priority, and I haven't seen any drawbacks to doing so, I have removed the grammar module's dependency on Ray.


AlpinDale:
Thanks for the PR! The RayRemote logits processor was added for multi-gpu scenarios where Ray is involved. Have you tested the PR with Ray enabled?

SwadicalRag:
Yes I have. I run a 2x 3090 system, and the performance was halved.

PR #278, opened 2024-02-28
fix logit bias logitproc

50h100a:
logit bias is a simple additive bias.
also, biases are floats, not longs.


mrsteyk:
Might want to fix this as well https://github.com/PygmalionAI/aphrodite-engine/blob/16615784b38f57fac196c19b5c048f41ea3848ec/aphrodite/endpoints/openai/api_server.py#L552-L557

PR #280, opened 2024-02-29
feat: add cubic sampling

AlpinDale:


StefanGliga:
It was a few missing unsqueezes, classic.
Out of precaution, I also added logic to guard against a sequence with smoothing factor set to zero(disabled) making it into this code. Quad/Cubic is one of the rarer sampler which doesn't have a value which is mathematically a no-op

PR #295, opened 2024-03-10
Fix typo in marlin kernel path.

sgsdxzy:
Fix build.


AlpinDale:
Thanks for catching that!

PR #296, opened 2024-03-10
Replace klite.embd with current version

Pyroserenus:
"you can make a PR, just copy paste the embed file from kcpp" alpin says


AlpinDale:
Thanks. Would you mind updating the kcpp versions in the API servers too?

Pyroserenus:
> Thanks. Would you mind updating the kcpp versions in the API servers too?

I understand some of those words.

PR #298, opened 2024-03-10
feat: INT8 KV Cache

AlpinDale:
INT8 kv cache with scales and zero points, needs calibration to export from the model. No throughput gain like FP8 E5m2, but also little to no accuracy loss as a result of the calibration. Only llama family of models are currently supported.

First, install additional requirements:
```sh
pip install fire datasets accelerate
```
Then generate state caches:
```sh
python aphrodite/kv_quant/calibrate.py --model meta-llama/Llama-2-7b-hf --calib_dataset wikitext2 --calib_samples 128 --calib_seqlen 4096 --work_dir kv_cache_states/llama-2-13b
```
Note that this uses almost 32GB of VRAM for Llama-2 7B.

Then finally, generate the scales and zero points:

```sh
python aphrodite/kv_quant/export_kv_params.py --work_dir kv_cache_states/llama-2-13b --kv_params_dir quant_params/llama-2-13b
```

You can then launch the API server with the extra args: `--kv-cache-dtype int8 --kv-quant-params-path quant_params/llama-2-7b`


AlpinDale:
Can't see to add AniZpZ as a co-author, weird.

Co-authored-by: aniz1905@gmail.com <zhangpeng156@meituan.com>

PR #305, opened 2024-03-11
feat: exllamav2 quantization

AlpinDale:
PR adds a functioning exllamav2 quantization scheme, huge thanks to @chu-tianxiang for figuring out the weight loading logic + vocab embedding changes.

TODO:
- [ ] Figure out tensor parallelism


AlpinDale:
TP issue may be due to kernels, will figure it out later.

PR #313, opened 2024-03-12
feat: add perplexity example

AlpinDale:
This PR adds an example to calculate model perplexity, this approach performs it by using `prompt_logprobs`:

- Generate text using a dataset and extract 1 prompt logprob per token.
- Calculate the mean of the logprobs.
- Calculate the exponent of the *negative* mean of the logprobs.
- Divide the result by 2.

The last step should technically be unneeded, but to match the perplexity results from llama.cpp, we seem to need the division by two. It still seems unreliable, as Llama-2 7B shows a lower ppx than Mistral 7B, and FP8 KV Cache doesn't change ppx at all.

### DIFF
--- a/examples/perplexity.py
+++ b/examples/perplexity.py
@@ -1,0 +1,51 @@
+import numpy as np
+from datasets import load_dataset
+from transformers import AutoTokenizer
+from aphrodite import LLM, SamplingParams
+
+# Load the wikitext2 dataset.
+dataset = load_dataset('wikitext', 'wikitext-2-raw-v1')
+
+# Get the first 2000 elements from the 'train' split.
+prompts = dataset['train']['text'][:2000]
+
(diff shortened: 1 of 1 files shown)


examples/perplexity.py
```diff
@@ -0,0 +1,51 @@
...
+dataset = load_dataset('wikitext', 'wikitext-2-raw-v1')
+
+# Get the first 2000 elements from the 'train' split.
+prompts = dataset['train']['text'][:2000]
+
+model_id = "mistralai/Mistral-7B-Instruct-v0.2"
+# Create a tokenizer.
+tokenizer = AutoTokenizer.from_pretrained(model_id)
+
+# Tokenize the prompts and discard or truncate any prompts longer than 2048 tokens.
```
Israel-Laguan, line 16: max_length is different below, maybe you meant 4096?

PR #332, opened 2024-03-16
feat: CMake Build System Generator

AlpinDale:
Ref. [Daniel Pfiefer's 2017 Talk](https://www.youtube.com/watch?v=bsXLMQ6WgIk) and [vllm#2830](https://github.com/vllm-project/vllm/pull/2830/)

This PR switches to a cmake build system. Much cleaner, and shortens build times.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ from torch.utils.cpp_extension import CUDA_HOME
 
 ROOT_DIR = os.path.dirname(__file__)
 
-MAIN_CUDA_VERSION = "12.1"
+MAIN_CUDA_VERSION = "12.2"
 
 def is_sccache_available() -> bool:
     return which("sccache") is not None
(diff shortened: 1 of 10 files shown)


setup.py
```diff
@@ -1,26 +1,164 @@
...
+
+            ext_target_name = remove_prefix(ext.name, "aphrodite.")
+            num_jobs, _ = self.compute_num_jobs()
+
+            build_args = [
+                '--build', '.', '--target', ext_target_name, '-j',
+                str(num_jobs)
+            ]
+
+            subprocess.check_call(['cmake', *build_args], cwd=self.build_temp)
```
nyxkrage, line 161: Is there potentially a better way to handle this, to let CMake handle all the extensions and scheduling, that way you wouldn't run out of things when each extension is close to being done when on machine with a high core count?

nyxkrage (commented):
The CMake specific stuff looks good to me, there's nothing that's jumping out as wrong

PR #338, opened 2024-03-17
docker-fixes

StefanDanielSchwarz:
@AlpinDale: With [Pull Request #327](https://github.com/PygmalionAI/aphrodite-engine/pull/327) you made some changes to "allow docker port and host to be changed", but that was already the case and these changes broke it because of how Docker works internally. So here's a fix for that to make the Docker stuff work again, and my explanations of what's changed (and why):

- The HOST must always be 0.0.0.0 since that's for listening on all interfaces *inside* the container, there's no other sensible option, as 127.0.0.1 would restrict it to localhost *inside* the container (preventing access from outside of Docker, and even from the host), and no other hostname/address would exist *inside* the container (since we don't have any other networks within the aphrodite-engine service). That's why I removed that variable again, reverting to the one and only appropriate value, 0.0.0.0.

- The PORT was already configurable before, as that variable determines which *Docker host port* the aphrodite-engine is exposed on. It only needs to change the left-side value of the ports directive in the Composefile, it shouldn't be changed *inside* the container as that has no effect on the host and the value has to be coordinated across files (envfile, Dockerfile, Composefile). Since you made 7860 the new default port instead of the 5000 we had before, I kept that change, but reverted/fixed the other instances.

- Also fixed a bug where the envfile introduced the KOBOLD_API variable but the entrypoint looks for just KOBOLD. Now it's KOBOLD_API for both and will work correctly.

- Finally, I moved NUMBA_CACHE_DIR from the entrypoint to the Dockerfile and made it a subdir of $HF_HOME. Makes more sense that way to me, but the only actual difference here is that it's now compatible with changing the HF_HOME inside the container, if we (or a user) ever want to override that.

Please build a new Docker image for all the changes to be applied correctly! Thanks!

And feel free to mention me with future Docker changes if you'd like me to look over them before merging. I do this stuff for a living and will happily support one of my favorite AI projects with my devops expertise. :)

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,11 +1,10 @@
 #!/bin/bash -e
 
-export NUMBA_CACHE_DIR="/tmp/numba_cache"
 echo 'Starting Aphrodite Engine API server...'
 
 CMD="python3 -m aphrodite.endpoints.openai.api_server
-             --host ${HOST:-0.0.0.0}
-             --port ${PORT:-7860}
+             --host 0.0.0.0
+             --port 7860
              --download-dir ${HF_HOME:?}/hub
              ${MODEL_NAME:+--model $MODEL_NAME}
              ${REVISION:+--revision $REVISION}
@@ -16,7 +15,7 @@ CMD="python3 -m aphrodite.endpoints.openai.api_server
              ${GPU_MEMORY_UTILIZATION:+--gpu-memory-utilization $GPU_MEMORY_UTILIZATION}
              ${QUANTIZATION:+--quantization $QUANTIZATION}
              ${ENFORCE_EAGER:+--enforce-eager}
-             ${KOBOLD:+--launch-kobold-api}
+             ${KOBOLD_API:+--launch-kobold-api}
              ${CMD_ADDITIONAL_ARGUMENTS}"
 
 # set umask to ensure group read / write at runtime
--- a/docker/.env
+++ b/docker/.env
@@ -5,8 +5,7 @@
 #GID=0
 #TZ=UTC
 
-#PORT=5000
-#HOST=0.0.0.0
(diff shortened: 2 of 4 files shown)


AlpinDale:
Thanks for the fix! I'm not very good with docker so I have to make do with GH co-pilot XD

docker/entrypoint.sh
```diff
@@ -16,7 +15,7 @@ CMD="python3 -m aphrodite.endpoints.openai.api_server
              ${GPU_MEMORY_UTILIZATION:+--gpu-memory-utilization $GPU_MEMORY_UTILIZATION}
              ${QUANTIZATION:+--quantization $QUANTIZATION}
              ${ENFORCE_EAGER:+--enforce-eager}
-             ${KOBOLD:+--launch-kobold-api}
```
AlpinDale, line 19: Good catch, can't believe this one slipped me...

AlpinDale (approved):

StefanDanielSchwarz:
> Thanks for the fix! I'm not very good with docker so I have to make do with GH co-pilot XD

Hehe, yeah, AI assistance is very helpful and I use it all the time, too – but it's usually just 80 % of the work, and it's most effective when you already know what to expect and where it tends to make mistakes. So, at least for now, it's not AI that will replace devs – but it'll be tough for juniors since seniors become so much more effective with AI augmentation.

> On a somewhat related note, do you think it's a good idea to add `hf_transfer` to requirements, and export `HF_HUB_ENABLE_HF_TRANSFER=1` in the docker env? It'd get us much faster downloads, but it's a bit unstable, so I'm not sure what to do here.

I like that, I use that setting myself, so having it inside the container would also be welcome. Didn't have any issues with it yet, so I'd say we enable it by default in the envfile.

PR #339, opened 2024-03-17
hf_transfer for faster downloads

StefanDanielSchwarz:
@AlpinDale As discussed in [Pull Request #338](https://github.com/PygmalionAI/aphrodite-engine/pull/338), I've added hf_transfer for faster downloads.

The official docs ([Environment variables](https://huggingface.co/docs/huggingface_hub/v0.21.4/package_reference/environment_variables#hfhubenablehftransfer)) list its limitations and state "Consequently, `hf_transfer` is not enabled by default in `huggingface_hub`." That's why I've added it to the envfile, but kept it commented out, so users can actively but easily choose to enable it.

Also changed HUGGING_FACE_HUB_TOKEN to HF_TOKEN in the envfile while I was at it. The docs ([Environment variables](https://huggingface.co/docs/huggingface_hub/v0.21.4/package_reference/environment_variables#deprecated-environment-variables)) state that HUGGING_FACE_HUB_TOKEN is deprecated and has been replaced with HF_TOKEN.

Note: Requires a Docker image rebuild since we're adding a new requirement that needs to be installed inside the container.

### DIFF
--- a/docker/.env
+++ b/docker/.env
@@ -1,4 +1,4 @@
-#HUGGING_FACE_HUB_TOKEN=<secret>
+#HF_TOKEN=<secret>
 
 #HF_CACHE=~/.cache/huggingface
 #UID=1000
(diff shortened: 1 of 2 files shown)


AlpinDale (approved):
Thanks for the PR!

PR #341, opened 2024-03-17
feat: CMake Build System

AlpinDale:
Trying #332 again.


setup.py
```diff
@@ -1,26 +1,164 @@
...
+            num_jobs = os.cpu_count()
+
+        nvcc_cuda_version = get_nvcc_cuda_version()
+        if nvcc_cuda_version >= Version("11.2"):
+            nvcc_threads = int(os.getenv("NVCC_THREADS", 8))
+            num_jobs = max(1, round(num_jobs / (nvcc_threads / 4)))
+        else:
+            nvcc_threads = None
+
+        return num_jobs, nvcc_threads
```
nyxkrage, line 64: the nvcc `-threads` options seemingly does nothing, and instead just passing the amount of jobs wanted as `-j X` to `cmake --build` seems to properly spawn the correct amount of nvcc/cicc/ptxas processes.
Should therefore be changed to 
```
    def compute_num_jobs(self):
        try:
            # os.sched_getaffinity() isn't universally available, so fall back
            # to os.cpu_count() if we get an error here.
            num_jobs = len(os.sched_getaffinity(0))
        except AttributeError:
            num_jobs = os.cpu_count()

        return os.getenv('MAX_JOBS', num_jobs)
```
so that it also supports settings amount of processes spawned at once with the `MAX_JOBS` environment variable

setup.py
```diff
@@ -1,26 +1,164 @@
...
+
+    # Perform cmake configuration for a single extension.
+    def configure(self, ext: CMakeExtension) -> None:
```
nyxkrage, line 67: This only gets run once due to the early return being hit on every subsequent extension. Therefor I suggest removing the `ext` parameter and then removing the `did_config` checks, changing outdir to `os.path.abspath(os.path.join(self.build_lib, "aphrodite"))` so that it stays the same

setup.py
```diff
@@ -1,26 +1,164 @@
...
+        if _install_hadamard():
+            cmake_args += ['-DAPHRODITE_INSTALL_HADAMARD_KERNELS=ON']
+
+        #
+        # Setup parallelism and build tool
+        #
+        num_jobs, nvcc_threads = self.compute_num_jobs()
+
+        if nvcc_threads:
+            cmake_args += ['-DNVCC_THREADS={}'.format(nvcc_threads)]
```
nyxkrage, line 122: Since nvcc_threads wont be used this can just be changed to just
```
num_jobs = self.compute_num_jobs()
```

setup.py
```diff
@@ -1,26 +1,164 @@
...
+
+            ext_target_name = remove_prefix(ext.name, "aphrodite.")
+            num_jobs, _ = self.compute_num_jobs()
+
+            build_args = [
+                '--build', '.', '--target', ext_target_name, '-j',
+                str(num_jobs)
+            ]
+
+            subprocess.check_call(['cmake', *build_args], cwd=self.build_temp)
```
nyxkrage, line 161: This loop can be removed and the contents can then be dedented, and for the build args the `--target ext_target_name` can be removed. That way CMake/ninja will handle all of the building/orchestrating of the builds for the extensions and you wont end up with having 12 jobs but only 4 things left to do at the end of each extension

CMakeLists.txt
```diff
@@ -0,0 +1,341 @@
...
+Query torch for additional GPU compilation flags for the given
+`APHRODITE_GPU_LANG`.
+The final set of arches is stored in `APHRODITE_GPU_FLAGS`.
+#]=========================================================================]
+get_torch_gpu_compiler_flags(APHRODITE_GPU_FLAGS ${APHRODITE_GPU_LANG})
+
+# Set nvcc parallelism.
+if(NVCC_THREADS AND APHRODITE_GPU_LANG STREQUAL "CUDA")
+  list(APPEND APHRODITE_GPU_FLAGS "--threads=${NVCC_THREADS}")
+endif()
```
nyxkrage, line 128: See comment in setup.py about nvcc --threads seemingly not doing anything, at least with the Ninja backend with CMake

nyxkrage (changes requested):
Ok, added some comments on improving parallelization of the build, and making it respect MAX_JOBS. With the changes I have commented it build and works properly on my machine just as on main

AlpinDale:
Actual cancer, trying to rebase this. I'll just remake the PR.

PR #343, opened 2024-03-19
feat: typical_p threshold sampling

AlpinDale:
PR adds a new hyperparameter to typical_p sampling, which scales the maximum threshold for positive deviations in typ_p. Credits to Suikamelon (@BugReporterZ ). Untested yet.


BugReporterZ:
Here is an explanation of what the modified code is supposed to do.

1. Calculate the range of candidate tokens as per original Typical_p algorithm;
2. Find the negative deviation $-D$ of the most likely token and invert its sign (let's call it $+D$);
3. Find the set of tokens having a deviation from $0$ to $+D$ in the original token distribution;
4. The final range of candidate tokens is the union of both sets.

Optionally, we can opt to scale $+D$ by a multiplier $n$; this has theoretical justifications, but results may vary depending on the model and use case. A safe value that always seems to be beneficial is 1.0. Values up to around 1.5~2.0 may be used but could seemingly occasionally lead to incoherence. 0.0 restores the original Typical-p behavior.

This modification has a much greater impact at low values of Typical-P than at high ones. From testing, good all-around typical-P values around 0.20-0.35 and a multiplier of 1.0 work very well.

aphrodite/modeling/layers/sampler.py
```diff
@@ -508,24 +509,45 @@ def _apply_epsilon_cutoff(
...
+    THRESHOLD = 1000
+
     shifted_logits = torch.log_softmax(logits, dim=-1)
-    probs = shifted_logits.exp()
+    probs = torch.exp(shifted_logits)
 
     neg_entropy = (probs * shifted_logits).nansum(dim=-1, keepdim=True)
+    # NOTE: We don't take the absolute value of the surprisal deviations
+    # This deviates from the original implementation
+    surprisal_deviations = neg_entropy - shifted_logits
```
BugReporterZ, line 525: I disagree with this change (or the intentions here). The modification in my original hack (not posted here) was intended to retain the basic behavior of Typical_P, which first sorts the surprisal deviations by their absolute value.

Only after this is done, then, using the signed surprisal deviations (copied into a different tensor before computing the absolute value for the other), you would obtain a second subset for extending the token selection as in the algorithm described in the explanation posted in the PR discussion.

BugReporterZ:
Hopefully this graphical explanation further clarifies how the modified algorithm is supposed to work.

image

BugReporterZ:
Further testing over the past few days has revealed that,

>[...] if the token having $-D$ deviation is an acceptable choice, then the one having $+D$ deviation should also be.

was actually _not_ a fair assumption and may too easily lead to incoherent text. While lower values could be used with good results, like $+D\over 2$, that is probably still too arbitrary.

A saner approach to this (while still retaining an additional hyperparameter over the original algorithm) would be scaling the negative deviations by a factor (Lambda) that defaults to 1.0 (same behavior as the original algorithm), before they are converted to absolute values in the code.

By doing this, the tokens corresponding to the scaled deviations would be de-emphasized, and the token selection progressively skewed toward rarer tokens the higher Lambda is.

Below is a simulated example using Tau=0.35 and Lambda in the 1.0~2.0 range. It can be seen how at Lambda > 1 the selection (yellow boxes) skews more toward rarer tokens:

image

- Advantages
   - The selection range scales up more progressively toward rarer tokens while retaining the original Typical-p behavior.
   - The modified algorithm is simpler than the originally-proposed modification.
   - No arbitrary theoretical assumptions made.
- Disadvantages
   - The oddly deterministic behavior of Typical-p at Tau=0.0 remains (the only token choice is the one with deviation=0, which is not the most likely token).
   - It would be desirable not to affect the deviation of control tokens like EOS, otherwise strange behavior may be observed (e.g. excessively long chat messages, etc). However this is a general problem that can affect other samplers as well.

I have code for the logic in `sampler.py`, but it's still based around the `main` branch. The portion for ignoring the first few tokens (UNK, EOS and BOS in Mistral and Llama tokenizers) could be ignored. Lambda would need to be plugged to an additional hyperparameter (which for the previous version in this PR branch was named Sigma).

```python
def _apply_typical_sampling(
    logits: torch.Tensor,
    typical_p: torch.Tensor,
) -> torch.Tensor:
    typ_p = torch.tensor(typical_p, dtype=logits.dtype, device=logits.device)
    shifted_logits = torch.log_softmax(logits, dim=-1)
    probs = shifted_logits.exp()

    # Tensor names and logic have been slightly altered to make the procedure a bit more consistent
    # with what the original Typical-P algorithm is intended to perform, thus clearer.
    surprises = shifted_logits.neg()
    conditional_entropy = (probs * surprises).nansum(dim=-1, keepdim=True)
    surprisal_deviations = surprises - conditional_entropy

    # Scale negative surprisal deviations by a scaling factor Lambda. This de-emphasizes
    # high-probability tokens. Lambda would ideally be an additional hyperparameter to typical-p.
    lambda_factor = 1.50
    lambda_mask = surprisal_deviations < 0

    # Don't affect special tokens (generally the first few tokens) by setting the mask to False
    # for them. This is an ugly hack; ideally we would want to identify such tokens more directly
    # as there is no guarantee that the first few tokens are special tokens or bytes (e.g. Qwen).
    tokens_to_ignore = 3
    lambda_mask[..., :tokens_to_ignore] = False

    # Actual scaling performed here.
    surprisal_deviations[lambda_mask] = surprisal_deviations[lambda_mask] * lambda_factor

    # From now on, the algorithm proceeds as in the original one, sorting tokens by absolute
    # surprisal deviations and picking them depending on their cumulative probability.
    surprisal_deviations = surprisal_deviations.abs()
    _, indices = torch.sort(surprisal_deviations)

    reordered_probs = probs.gather(-1, indices)
    typ_mask_sorted = reordered_probs.cumsum(dim=-1) >= typ_p.unsqueeze(dim=1)

    min_tokens_to_keep = 1
    # Keep at least min_tokens_to_keep
    typ_mask_sorted[..., :min_tokens_to_keep] = 0

    typ_mask = typ_mask_sorted.scatter(1, indices, typ_mask_sorted)
    logits[typ_mask] = -float("inf")
    return logits
```

BugReporterZ:
A different (untested) strategy with minimal modifications from the above could be, instead of scaling the positive deviations by a Lambda factor, shifting the entire deviations by a Delta value.

image

Possible advantage: the same control can be used for skewing the distribution toward more probable tokens.

PR #349, opened 2024-03-21
[WIP] feat: add grok-1 support

AlpinDale:
This [WIP] PR adds support for the [grok-1](https://github.com/xai-org/grok-1) model.


AlpinDale:
This may be kind of pointless.

PR #358, opened 2024-03-25
Minor Samplers Overhaul

50h100a:
Overhauled SamplingTensors construction.
Fix multiple bugs in sampler flags.
Restored the functioning logitproc format.
Fixed sudden NaNs in quadratic smoothing.
Rewrote Mirostat to work with seeds and other samplers.
Removed branches from some samplers.

This supercedes the previous miro PR I had.

Who's on deck for the guidance/grammar things? At least one of them is using a backwards logitproc formulation, might need a wrapper if a proper tensor-based edit isn't in the cards.

Still investigating a potential perf regression in output tokens. TBD.


50h100a:
Tired of playing merge tennis.

I suspect there may be issues with outlines/grammar, but I am unable to find an end-to-end test for them.
Similarly, I can't test the Triton seeding stuff, but it isn't enabled anyway.

PR #363, opened 2024-03-30
feat: add embeddings endpoint to openai rest-api server.

IggoOnCode:
This PR adds an /v1/embeddings endpoint to aphrodite.endpoints.openai.api_server with the goal of making aphrodite-engine usable with software that uses the same OpenAI-baseurl for embeddings and inference (for example CrewAI).

The embeddings code itself is taken from text-generation-webui. It uses sentence-transformer to create the embeddings.

As there are currently no contribution guidelines, please provide me with some direction what would be needed to make this code elegible for merging.


AlpinDale:
Thanks for the contribution! I can review this PR after the holidays, but for now, would you mind adding attributions to textgen webui for the embedding code?

We should probably add contribution guidelines, yes. Another matter to think about. Would be good if we can get simple test units in for this feature.

IggoOnCode:
@AlpinDale Thank you for the fast initial feedback. Have a nice holiday!

IggoOnCode:
I had problems getting the test suite to run, so I couldn't do real unit tests yet. I added a very minimal functional test to at least have some coverage.

Is there some documentation regarding the tests? Just doing `./runtime.sh pytest tests/path/to/testfile` only worked in very few testcases. `./runtime.sh pytest` failed at conftest.py already.

aphrodite/endpoints/openai/embeddings.py
```diff
@@ -0,0 +1,137 @@
...
+    '''
+    using 'lazy loading' to avoid circular import
+    so this function will be executed only once
+    '''
+    global embeddings_params_initialized
+    if not embeddings_params_initialized:
+
+        global st_model, embeddings_model, embeddings_device
+
+        st_model = os.environ.get("OPENEDAI_EMBEDDING_MODEL",
```
AlpinDale, line 28: Is the typo with `OPENEDAI` here intentional?

AlpinDale:
@IggoOnCode Thanks for the contribution. I made a few QoL changes:

- Allow passing arbitrary text embedding models via the API
- Protect the route with an API key
- Add pydantic classes for request and output objects (mainly for swagger UI)

I'll figure out the tests later. I think it's ready for merge as-is. For a future PR, we could potentially extract embeddings directly from the loaded model.

aphrodite/endpoints/openai/embeddings.py
IggoOnCode: No

AlpinDale (approved):
LGTM

IggoOnCode:
Thank you for taking up the code and especially for giving it the last touches! It's very much appreciated.

PR #377, opened 2024-04-02
chore: make metadata a dataclass

AlpinDale:


AlpinDale:
approval done in private

PR #381, opened 2024-04-03
Support arbitrary model in GGUF.

sgsdxzy:
Offline conversion should work for any model if the `config.json` and tokenizer from the base repo is copied to output dir.
Online conversion is still limited to llama because only llama config is supported.


AlpinDale (approved):
LGTM, please resolve conflicts then we can merge.

PR #412, opened 2024-04-16
Fix max_num_batched_tokens for chunked_prefill.

sgsdxzy:


requirements-common.txt
```diff
@@ -4,6 +4,7 @@ psutil
 sentencepiece
 numpy
 transformers == 4.39.3 # needed for llava
+accelerate
```
AlpinDale, line 7: Can we put this in requirements-cuda.txt instead? GGUF support at the moment is CUDA only. We should also lazily import accelerate in hf_downloader.py, so it's not imported too early if GGUF isn't being used.

sgsdxzy: I removed accelerate completely.

AlpinDale (approved):

PR #419, opened 2024-04-17
[WIP] feat: Mamba support

AlpinDale:
Support for mamba, including the hybrid architectures jamba and zamba.


AlpinDale:
It works, but it seems to produce slightly coherent text. Needs to be investigated.

PR #423, opened 2024-04-19
Split the exl2 weight ASAP.

sgsdxzy:
Reduce vram usage during exl2 model loading.

### DIFF
--- a/aphrodite/modeling/layers/linear.py
+++ b/aphrodite/modeling/layers/linear.py
@@ -26,6 +26,68 @@ def adjust_marlin_shard(param, shard_size, shard_offset):
     return shard_size * marlin_tile_size, shard_offset * marlin_tile_size
 
 
+# Split the exl2 weight at group boundary
+def post_init_exl2(layer):
+    q_groups = layer.q_groups
+    num_groups = q_groups.shape[0] // 2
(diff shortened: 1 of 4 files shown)


sgsdxzy:
The outer `model.named_parameters()` loop during loading still keeps a reference to weights so we have to force release memory by setting `.data=torch.empty(0, device="cpu")`.
`q_invperm` is not used later in the kernel, probably we should remove it too after loading.

sgsdxzy:
Sharding on CPU then copying the contiguous tensor reduces GPU vram bubble, increasing GPU blocks further (36k->41k ctx for me).

aphrodite/modeling/layers/linear.py
```diff
@@ -26,6 +26,69 @@ def adjust_marlin_shard(param, shard_size, shard_offset):
     return shard_size * marlin_tile_size, shard_offset * marlin_tile_size
 
 
+# Split the exl2 weight at group boundary
```
AlpinDale, line 29: Why is this moved to this file?

sgsdxzy: It would be a circular import otherwise.

AlpinDale (approved):
Thanks for the PR! This should fix SQ+ as well.

PR #424, opened 2024-04-22
Fix: kobold api /tokencount

Krovius:
Existing source attempts to pass a string as 'Prompt' to the tokenizer and performs a len() on a dict, returning the number of members and not the token count.


AlpinDale (approved):
LGTM, thanks for the fix.

PR #444, opened 2024-04-29
fix: restore backwards compatibility with sm_60 (P100 and GP100)

### LINKED ISSUE #413: [Feature]: Is there a reason CUDA 6.1 is the minimum? Would CUDA 6.0 on the P100 not work?
### 🚀 The feature, motivation and pitch

In the setup.py it checked for CUDA 6.1 as a minimum and that requirement is also stated in the readme. Is there a technical reason CUDA 6.0 is not supported? Is it for INT8 support? 

I ask this because there is nothing inherently stopping VLLM which aphrodite is forked from, from working with CUDA 6.0 on the Tesla P100 cards. As can be seen in this dicussion: https://github.com/vllm-project/vllm/issues/963#issuecomment-1863147987

```
if _is_cuda() and not compute_capabilities:
    # If TORCH_CUDA_ARCH_LIST is not defined or empty, target all available
    # GPUs on the current machine.
    device_count = torch.cuda.device_count()
    for i in range(device_count):
        major, minor = torch.cuda.get_device_capability(i)
        if major < 6 or (major == 6 and minor < 1):
            raise RuntimeError(
                "GPUs with compute capability below 6.1 are not supported.")
        compute_capabilities.add(f"{major}.{minor}")
```

### Alternatives

_No response_

### Additional context

_No response_

AlpinDale:
Looks like sm_60 can't do dot product with 4 elements. There's a better solution, but this will work for now. I may implement an sm_60-only dot product function here.

resolves #413

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@ include(${CMAKE_CURRENT_LIST_DIR}/cmake/utils.cmake)
 set(PYTHON_SUPPORTED_VERSIONS "3.8" "3.9" "3.10" "3.11")
 
 # Supported NVIDIA architectures.
-set(CUDA_SUPPORTED_ARCHS "6.1;7.0;7.5;8.0;8.6;8.9;9.0")
+set(CUDA_SUPPORTED_ARCHS "6.0;6.1;7.0;7.5;8.0;8.6;8.9;9.0")
 
 # Supported AMD GPU architectures.
 set(HIP_SUPPORTED_ARCHS "gfx906;gfx908;gfx90a;gfx940;gfx941;gfx942;gfx1030;gfx1100")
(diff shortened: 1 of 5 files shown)


dirkson:
I was able to build and run this on 1x and 4x P100's more or less flawlessly with the provided runtime. I tested turboderp's Llama 3 8B exl's and MaziyarPanahi's 8B and 70B gptq.

There was an issue with --context-shift , which is apparently due to Triton currently being limited to Cuda 7.0. The GPTQ's endlessly generated, but that's a known issue with the models tested, so I suspect it was unrelated.

Otherwise I really wasn't able to find any issues. Everything I tried just worked in my setup, or gave reasonable error messages about requiring higher Cuda releases.

Performance, once I adjusted to the way Aphrodite reports performance, was quite pretty solid - The 70B model on -tp 4 ran at 12t/s, up from 3.5 for tabbyapi. The 8B model ran at 50t/s on a single P100, or 25t/s on multiple. It ran at around 40t/s on tabbyapi.

PR #449, opened 2024-04-30
Fix minor bugs in outlines and lmfe.

sgsdxzy:
https://github.com/vllm-project/vllm/pull/4189
Bump lmfe version to 0.9.8 to fix various issues.


sgsdxzy:
Updated, also fixes https://github.com/noamgat/lm-format-enforcer/issues/94

PR #450, opened 2024-05-03
feat: add batch tokenization endpoint along with option for no token ids

ahme-dev:


aphrodite/endpoints/openai/api_server.py
```diff
@@ -91,6 +91,19 @@ async def tokenize(request: Request,
     return JSONResponse(content=tokenized)
 
 
+@router.post("/v1/tokenize-batch")
+@router.post("/v1/token/encode-batch")
+async def tokenize_batch(request: Request,
+                         prompts: List[Prompt],
+                         no_ids: Optional[bool] = False,
+                         x_api_key: Optional[str] = Header(None)):
+    tokenized_responses = []
```
Krakenos, line 100: Instead of:
```python
 tokenized_responses = []
 for p in prompts:
    tokenized = await openai_serving_chat.tokenize(p, noIds=no_ids)
    tokenized_responses.append(tokenized)
return JSONResponse(content=tokenized_responses)
```
It should be better to do something like:
```python
tasks = [asyncio.create_task(openai_serving_chat.tokenize(p, noIds=no_ids) for p in prompts]
tokenized_responses = asyncio.gather(*tasks)
return JSONResponse(content=tokenized_responses)
```
Because the first one will await and go one by one for each prompt. The latter will execute all the tasks for tokenization concurrently.

ahme-dev: So it's similar to JS' promise.all, that's much better. Thanks

PR #454, opened 2024-05-07
Refactor: Quantization

AlpinDale:
Some light refactors to isolate the quantization code from the regular stuff. Also lets users disable compilation for quant kernels by passing the `APHRODITE_INSTALL_QUANT_KERNELS=0` env variable.


setup.py
```diff
@@ -217,6 +220,18 @@ def _is_cpu() -> bool:
     return APHRODITE_TARGET_DEVICE == "cpu"
 
 
+def _install_quants() -> bool:
+    install_quants = bool(
+        int(os.getenv("APHRODITE_INSTALL_QUANT_KERNELS", "1")))
+    device_count = torch.cuda.device_count()
+    for i in range(device_count):
+        major, minor = torch.cuda.get_device_capability(i)
+        if major <= 6:
```
sgsdxzy, line 229: I think GPTQ is supported on cc 6.x, do you really want to disable it.

setup.py
```diff
@@ -235,7 +250,7 @@ def _install_hadamard() -> bool:
     device_count = torch.cuda.device_count()
     for i in range(device_count):
         major, minor = torch.cuda.get_device_capability(i)
-        if major <= 6:
+        if major <= 7:
```
sgsdxzy, line 253: hadamard works on Turing (7.5) but I haven't tested quip#

sgsdxzy (commented):
I think it would be much better to check `HAS_QUANTS` in each `QuantizationConfig` subclass' `__init__` rather than `apply_weights`, it errs earlier and reduces unnecessary changes to code.

AlpinDale:
> I think it would be much better to check `HAS_QUANTS` in each `QuantizationConfig` subclass' `__init__` rather than `apply_weights`, it errs earlier and reduces unnecessary changes to code.

Good idea, I'll change it to that.

PR #459, opened 2024-05-08
Fix recursion errors with large amounts of blocks

thomas-xin:
This is a small fix to processing/block/common.py in the scenario where the amount of blocks exceeds Python's recursion limit, causing the engine to error upon trying to retrieve the linked blocks as a list. The inline `recurse` function is replaced by a while loop traversing the blocks without recursion.

Engine should function as normal on such blocks with this fix applied.

### DIFF
--- a/aphrodite/processing/block/common.py
+++ b/aphrodite/processing/block/common.py
@@ -161,11 +161,10 @@ def get_all_blocks_recursively(last_block: Block) -> List[Block]:
             appear.
     """
 
-    def recurse(block: Block, lst: List[Block]) -> None:
-        if block.prev_block is not None:
-            recurse(block.prev_block, lst)
-        lst.append(block)
-
-    all_blocks = []
-    recurse(last_block, all_blocks)
+    block = last_block
+    all_blocks = [block]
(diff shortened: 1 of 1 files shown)


AlpinDale:
I'm debating whether it's better to increase the recursion limit or go with this fix, but I can't say for certain unless I can reproduce the issue (for benchmarking and correctness tests). Can you provide reproduction steps?

thomas-xin:
> I'm debating whether it's better to increase the recursion limit or go with this fix, but I can't say for certain unless I can reproduce the issue (for benchmarking and correctness tests). Can you provide reproduction steps?

I've encountered this most when using chunked prefill or the v2 block manager with cohere models (although I'm not sure if other models run into this issue too), but it appears to happen at around 60000+ context size. With the fix here I can confirm it working with 120k context.

Upping the recursion limit with python is possible but a little iffy because it introduces the possibility of stack overflow, which is possibly even less desirable here. Is there a more specific reason the recursive traversal is used here as opposed to iterative? Apart from perhaps slightly easier to read code

PR #536, opened 2024-07-15
Update Kobold Lite Embed

Pyroserenus:
Syncs the kobold lite embed and disables certain features that aphrodite cannot currently use. KoboldCPP impersonation version has not been incremented as no new features need to be enabled.

### DIFF
--- a/aphrodite/endpoints/kobold/klite.embd
+++ b/aphrodite/endpoints/kobold/klite.embd
@@ -5156,19 +5156,19 @@ Current version indicated by LITEVER below.
 	}
 	function is_using_kcpp_with_added_memory()
 	{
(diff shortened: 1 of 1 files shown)


LostRuins:
Should list the main changes or better yet separate those into a different comnit... so future devs will be able to replicate this effort. As lite is bound to keep evolving and changing. From first glance I can't actually tell what was modified

Pyroserenus:
> Should list the main changes or better yet separate those into a different comnit... so future devs will be able to replicate this effort. As lite is bound to keep evolving and changing. From first glance I can't actually tell what features was modified or missing

It is a different commit, there's two commits with this PR, the first replaces the klite embed with the current one, the second adjusts non functional features to require koboldcpp impersonation at 1.71

LostRuins:
True, fair enough

PR #538, opened 2024-07-21
Remove `$` from the shell code blocks in README

matthusby:
When using the copy button this also copies the `$` so when it is pasted the commands are invalid.


AlpinDale (approved):
Good call, thanks!

PR #560, opened 2024-07-29
feat: TorchAO floating point quantization

intervitens:
This PR adds a custom floating point quantization method powered by [TorchAO](https://github.com/pytorch/ao), which achieves a high throughput, thanks to the optimized [fp6_llm](https://github.com/usyd-fsalab/fp6_llm) kernel.

Use `-q torchao --torchao-fp-bits 6` to load an FP16 model and convert it at runtime to fp6_e2m3 format.

Using with tensor parallelism currently results in degraded outputs.

### DIFF
--- a/aphrodite/quantization/gptq.py
+++ b/aphrodite/quantization/gptq.py
@@ -203,13 +203,7 @@ class GPTQLinearMethod(LinearMethodBase):
 
         layer.exllama_state = exllama_state
 
-    def apply(self,
-              layer: torch.nn.Module,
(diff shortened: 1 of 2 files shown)


intervitens:
Added splitK calculation, speeds up GSM8k benchmark run with bs=32 on a single 3090 ti from 8:47 to 7:57

AlpinDale:
I couldn't cleanly merge my changes with your latest commit, I'll add the SplitK myself in a bit (this is still a WIP)

PR #571, opened 2024-08-08
feat: initial XPU support via intel_extension_for_pytorch

AlpinDale:


AlpinDale:
GQA doesn't work yet due to limitations upstream.

PR #594, opened 2024-08-23
feat: Asymmetric Tensor Parallel

AlpinDale:
To run models with tensor parallel sizes where their number of attention heads is not divisible by the GPU count.


AlpinDale:
Seems to work well, with no throughput penalties.

PR #644, opened 2024-09-07
quadratic sampling: separate diff from logits to filter out NaNs.

50h100a:
Infs lead to NaNs in the logits which make samplers very sad.


AlpinDale (approved):
thanks for le fix

PR #648, opened 2024-09-07
bringing dynatemp back

50h100a:
like it never even left

### DIFF
--- a/aphrodite/modeling/sampling_metadata.py
+++ b/aphrodite/modeling/sampling_metadata.py
@@ -308,6 +308,9 @@ class SamplingTensors:
     """Tensors for sampling."""
 
     temperatures: torch.Tensor
+    dynatemp_mins: torch.Tensor
+    dynatemp_maxs: torch.Tensor
+    dynatemp_exps: torch.Tensor
     top_ps: torch.Tensor
     top_ks: torch.Tensor
     top_as: torch.Tensor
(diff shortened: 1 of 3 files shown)


aphrodite/modeling/sampling_metadata.py
```diff
@@ -683,7 +720,7 @@ def from_lists(cls, temperatures: List[float], top_ps: List[float],
 
     @staticmethod
     def _get_sequence_seeds(
-        seed: int,
```
AlpinDale, line 686: What's the reason for this?

AlpinDale:
Thanks for bringing it back, can you pipe it in the API server too?

aphrodite/modeling/sampling_metadata.py
50h100a: It's just correcting the typehint.

`seed` comes from an `int|None` source, and this func has explicit behavior for the `None` case.

50h100a:
> Thanks for bringing it back, can you pipe it in the API server too?
Which? I tested it with the OpenAI endpoint, it seems to function there.

AlpinDale:
> > Thanks for bringing it back, can you pipe it in the API server too?
> 
> Which? I tested it with the OpenAI endpoint, it seems to function there.

The API ignores any field that isn't present. You'll need to add it for the `ChatCompletionRequest` and `CompletionRequest` classes in the OAI server's procotol.

50h100a:
No wonder it ran fine.

PR #760, opened 2024-09-25
chore: update klite.embd

eltociear:
pseduo -> pseudo


AlpinDale:
This file is copied from [koboldcpp](https://github.com/LostRuins/koboldcpp/blob/concedo/klite.embd). I recommend sending a patch there. We'll eventually pull the fix once we update the klite embed file.

PR #764, opened 2024-09-27
Fix for a crash from token bans

Pyroserenus:
Fixes a crash from token bans during batched inference. Has been running stable for the past few hours and token bans seem to still work.


AlpinDale:
Please run `pip install -r requirements-dev.txt` then `./formatting.sh` to fix linting issues.

Pyroserenus:
linting issues rectified.

PR #775, opened 2024-10-14
Make amd usable

Naomiusearch:
Seems like profiling peak memory is broken/really slow for GPTQ, FP16 of LLAMA-3 8b works just fine


Naomiusearch:
GPTQ seems to work now for now reason(?)

docs/pages/installation/installation-rocm.md
```diff
@@ -72,6 +72,8 @@ Finally, build Aphrodite:
 git clone https://github.com/PygmalionAI/aphrodite-engine.git
 cd aphrodite-engine
 
+chmod +x ./amdpatch.sh
+./amdpatch.sh
```
AlpinDale, line 76: Can we add this to the Dockerfile as well?

AlpinDale: I may do it in a follow-up PR.

Naomiusearch: Not sure, cause I don't use docker. Though probably yes

PR #785, opened 2024-10-21
Stream models rather than load them completely into RAM.

50h100a:
Change only applied to LLaMA, and only when using DefaultModelLoader.

Requestin' inspection to see if this is practical to do everywhere.


BBC-Esq:
Is this kind of like AirLLM?

AlpinDale:
> Is this kind of like AirLLM?

No, this only applies to the model weight loading stage, to avoid excessive CPU RAM usage.

PR #814, opened 2024-11-13
fix: temperature issues

50h100a:
Temperature was broken because of the `.to(torch.float)` upcast making a copy and thus not modifying `logits` inline.

The rest addresses the ultra-low-temp overflow that the upcast was trying to fix in the first place.

For discussion:
1. This _should_ be safe. Is it? A sampler relying on exact logit magnitudes seems insane to me.
2. I'm only partly convinced by the [-inf, 1] range. Using [-inf, 0] would drop from 75% to 50% of available values, but simultaneously places the very highest precision at the top logits.

### DIFF
--- a/aphrodite/modeling/layers/sampler.py
+++ b/aphrodite/modeling/layers/sampler.py
@@ -318,13 +318,18 @@ def _apply_temperatures(
     normalized_entropies = dynatemp_entropies.div_(dynatemp_max_entropies)
     dyn_temp = (dynatemp_mins + (dynatemp_maxs - dynatemp_mins) *
                 normalized_entropies.pow_(dynatemp_exps))
-
-    temperatures[dynatemp_mask] = dyn_temp
-    temperatures[temperatures <= 0.0] = 1e-5
+  
+    # There isn't a "safe" temperature range for fp16 logits.
+    # This value was chosen because 1/2e-5 is just under the 65k fp16 max,
+    # meaning that this temperature well-uses the fp16 space after the offset.
+    temperatures[temperatures <= 2e-5] = 2e-5
+  
+    # To prevent saturation of top logits, we shift the range to [-inf, 1]
+    # Why align to 1, instead of 0? Because [0, 1] holds 25% of all floats.
+    # Why mask? So we aren't potentially discarding data in milder temps.
+    low_temps = temperatures < 0.1
+    logits[low_temps] -= logits.max(dim=-1, keepdim=True).values[low_temps] - 1
     logits.div_(temperatures.unsqueeze(dim=1))
-    # There really isn't a "safe" temperature range for fp16 logits.
-    # So just sanitize them afterwards! -inf is fine, it's +inf that's shonky.
-    logits[logits.isposinf()] = torch.finfo(torch.float16).max
 
 
 def _apply_token_bans(logits: torch.Tensor,
--- a/aphrodite/modeling/layers/sampler.py
+++ b/aphrodite/modeling/layers/sampler.py
@@ -318,6 +318,7 @@ def _apply_temperatures(
     normalized_entropies = dynatemp_entropies.div_(dynatemp_max_entropies)
     dyn_temp = (dynatemp_mins + (dynatemp_maxs - dynatemp_mins) *
                 normalized_entropies.pow_(dynatemp_exps))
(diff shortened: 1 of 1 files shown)


aphrodite/modeling/layers/sampler.py
```diff
@@ -318,12 +318,18 @@ def _apply_temperatures(
...
     temperatures[dynatemp_mask] = dyn_temp
-    temperatures[temperatures <= 0.0] = 1.0
-    # Use float32 to apply temp.
-    # Use in-place division to avoid creating a new tensor.
-    logits = logits.to(torch.float)
+  
+    # There isn't a "safe" temperature range for fp16 logits.
+    # This value was chosen because 1/2e-5 is just under the 65k fp16 max,
+    # meaning that this temperature well-uses the fp16 space after the offset.
+    temperatures[temperatures <= 2e-5] = 2e-5
```
AlpinDale, line 326: Instead of hardcoding it here, can we document it as a constant (e.g. `_MIN_TEMPERATURE = 2e-5`) at the top of the module?

aphrodite/modeling/layers/sampler.py
```diff
@@ -318,12 +318,18 @@ def _apply_temperatures(
...
+  
+    # To prevent saturation of top logits, we shift the range to [-inf, 1]
+    # Why align to 1, instead of 0? Because [0, 1] holds 25% of all floats.
+    # Why mask? So we aren't potentially discarding data in milder temps.
+    low_temps = temperatures < 0.1
```
AlpinDale, line 331: Any reason for choosing 0.1?

aphrodite/modeling/layers/sampler.py
```diff
@@ -318,12 +318,18 @@ def _apply_temperatures(
     normalized_entropies = dynatemp_entropies.div_(dynatemp_max_entropies)
```
AlpinDale, line 318: Unrelated to the PR, but I wonder if there might be cases where `dynatemp_max_entropies` would contain zeros (possible if all logits are -inf). This would end up producing NaNs.

aphrodite/modeling/layers/sampler.py
```diff
@@ -318,12 +318,18 @@ def _apply_temperatures(
...
+    # There isn't a "safe" temperature range for fp16 logits.
+    # This value was chosen because 1/2e-5 is just under the 65k fp16 max,
+    # meaning that this temperature well-uses the fp16 space after the offset.
+    temperatures[temperatures <= 2e-5] = 2e-5
+  
+    # To prevent saturation of top logits, we shift the range to [-inf, 1]
+    # Why align to 1, instead of 0? Because [0, 1] holds 25% of all floats.
+    # Why mask? So we aren't potentially discarding data in milder temps.
+    low_temps = temperatures < 0.1
+    logits[low_temps] -= logits.max(dim=-1, keepdim=True).values[low_temps] - 1
```
AlpinDale, line 332: Any performance impacts here? Considering the range shifting adds an extra max op and indexing for low temps. I can run a benchmark later if you can't

AlpinDale (changes requested):

aphrodite/modeling/layers/sampler.py
50h100a: That makes sense.

aphrodite/modeling/layers/sampler.py
50h100a: It's an arbitrary choice, simply trying to avoid touching the logits if there's no compelling reason to.
It translates into a 10x multiplier to logits, which permits values up to ~6000 before you run into trouble.

aphrodite/modeling/layers/sampler.py
50h100a: I believe it can, though practically speaking I would only expect this to be an issue with temperature_last.

(just tested, yes, yes, it can. easily fixed, though.)

aphrodite/modeling/layers/sampler.py
50h100a: No idea, but historically I've never seen sampling take up more than a percent or two of realistic workloads.
Also, keep in mind, almost every other sampler does a full _sort_ on the logits, let alone a mere O(N) `max`.

AlpinDale (approved):
LGTM

PR #825, opened 2024-11-19
feat: implement top-nsigma sampling method

AlpinDale:
This PR adds the top-nσ sampling method from the paper ["Top-nσ: Not All Logits Are You Need"](https://arxiv.org/abs/2411.07641).

Top-nσ seems to be a novel sampling method that operates directly on pre-softmax logits by leveraging statistical properties of the distribution. Instead of using complex probability manipulations like top-p/top-k, it filters tokens based on their distance from the maximum logit in terms of standard deviations.

I haven't tested it much, but values should be between 0.0 (disabled) and 2.0, probably.

TODO:

- [ ] See if this sampler should disable all other truncation samplers (top-p/k/a, min_p)
- [ ] Write tests

### DIFF
--- a/aphrodite/common/sampling_params.py
+++ b/aphrodite/common/sampling_params.py
@@ -375,7 +375,7 @@ class SamplingParams(
             raise ValueError(
                 "xtc_probability must be in [0, 1], got "
                 f"{self.xtc_probability}.")
-        if not self.nsigma <= 0.0:
+        if self.nsigma <= 0.0:
             raise ValueError(
                 "nsigma must be non-negative, got "
                 f"{self.nsigma}.")
--- a/aphrodite/common/sampling_params.py
+++ b/aphrodite/common/sampling_params.py
@@ -375,7 +375,7 @@ class SamplingParams(
             raise ValueError(
                 "xtc_probability must be in [0, 1], got "
                 f"{self.xtc_probability}.")
-        if self.nsigma <= 0.0:
+        if self.nsigma < 0.0:
             raise ValueError(
                 "nsigma must be non-negative, got "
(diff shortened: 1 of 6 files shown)


AlpinDale:
@tomorrowdawn hi! I was wondering if it's a good idea to disable alphabet sampling (and other truncators) when nsigma is being used? We used to do this for mirostat as well, but it would disable all other samplers instead of just truncators.

Tomorrowdawn:
It's an honor to see this PR! nsigma and top-p/min-p are definitely conflicting and thus not recommended. While it doesn't conflict with top-k, k typically only reduces the number of available tokens. I'm not sure what the practical requirements are, but technically this is how it works.

AlpinDale:
The current order is penalties -> temperature -> top-n-sigma -> alphabet. We also allow users to perform temperature sampling last if they wish. Do you recommend moving it to before or after temperature?

I had some users test this sampler offline, and they noted it performs better when temperature is applied before top-nsigma.

Tomorrowdawn:
It is same before or after temperature scaling, see the temperature invariance [Sec 3.2](https://arxiv.org/pdf/2411.07641). 

For the concern of quality: I guess the best choice is to let the user decide. All these methods are "denoising" the distribution, so the final result is bounded by *the most strict* sampler. At worst, it just tends toward greedy decoding.

AlpinDale:
I suppose you're right. I'll add sampler order soon so the users can control how this happens. Merging for now as it passes all tests.

AlpinDale:
Whoops, missed up the git command, but it's in PygmalionAI/aphrodite-engine@22427602ebd45652a0ea323dad3959430c6db352

PR #827, opened 2024-11-20
feat: Add DRY (Do not Repeat Yourself) sampling

selalipop:
A direct port of https://github.com/oobabooga/text-generation-webui/pull/5677/files

Key changes:
- Add DRY parameters to OpenAI models
- Add DRY parameters to SamplingParams
- Add DRY tensors to SamplingTensors
- Implement _apply_dry() logits processor

### DIFF
--- a/aphrodite/common/sampling_params.py
+++ b/aphrodite/common/sampling_params.py
@@ -153,6 +153,21 @@ class SamplingParams(
             (max_logit - nsgima * std_dev) are filtered out. Higher values
             (e.g. 3.0) keep more tokens, lower values (e.g. 1.0) are more
             selective. Must be positive. 0 to disable.
+        dry_multiplier: Float that controls the magnitude of the DRY sampling penalty.
+            Higher values create stronger penalties against repetition. The penalty
+            is multiplied by this value before being applied. Must be non-negative.
+            0 disables the sampler.
+        dry_base: Base for the exponential growth of the DRY sampling penalty.
+            Controls how quickly the penalty increases with longer repeated sequences.
+            Must be greater than 1. Higher values (e.g. 2.0) create more aggressive
+            penalties for longer repetitions. Defaults to 1.75.
+        dry_allowed_length: Maximum number of tokens that can be repeated without
+            incurring a DRY sampling penalty. Sequences longer than this will be
+            penalized exponentially. Must be at least 1. Defaults to 2.
+        dry_sequence_breaker_ids: List of token IDs that stop
+            the matching of repeated content. These tokens will break up the input
+            into sections where repetition is evaluated separately. Common examples
+            are newlines, quotes, and other structural tokens. Defaults to None.
     """
 
     n: int = 1
@@ -199,7 +214,10 @@ class SamplingParams(
     xtc_threshold: float = 0.1
     xtc_probability: float = 0
     nsigma: float = 0.0
-
+    dry_multiplier: float = 0.0
+    dry_base: float = 1.75
+    dry_allowed_length: int = 2
+    dry_sequence_breaker_ids: List[float] = []
     # The below fields are not supposed to be used as an input.
     # They are set in post_init.
     output_text_buffer_length: int = 0
@@ -246,6 +264,10 @@ class SamplingParams(
         "xtc_threshold": 0.1,
         "xtc_probability": 0,
         "nsigma": 0.0,
+        "dry_multiplier": 0.0,
+        "dry_base": 1.75,
+        "dry_allowed_length": 2,
+        "dry_sequence_breaker_ids": [],
     }
 
     def __post_init__(self) -> None:
@@ -379,7 +401,18 @@ class SamplingParams(
             raise ValueError(
                 "nsigma must be non-negative, got "
                 f"{self.nsigma}.")
-            
+        if self.dry_multiplier < 0.0:
+            raise ValueError(
+                "dry_multiplier must be non-negative, got "
+                f"{self.dry_multiplier}.")
+        if self.dry_base <= 1.0:
+            raise ValueError(
+                "dry_base must be greater than 1, got "
+                f"{self.dry_base}.")
+        if self.dry_allowed_length < 0:
+            raise ValueError(
+                "dry_allowed_length must be non-negative, got "
+                f"{self.dry_allowed_length}.")    
 
     def _verify_beam_search(self) -> None:
         if self.best_of == 1:
--- a/aphrodite/common/sampling_params.py
+++ b/aphrodite/common/sampling_params.py
@@ -219,7 +219,7 @@ class SamplingParams(
     dry_multiplier: float = 0.0
     dry_base: float = 1.75
     dry_allowed_length: int = 2
-    dry_sequence_breaker_ids: List[float] = []
+    dry_sequence_breaker_ids: List[int] = []
     # The below fields are not supposed to be used as an input.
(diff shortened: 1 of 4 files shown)


aphrodite/modeling/sampling_metadata.py
```diff
@@ -728,6 +747,31 @@ def from_lists(cls, temperatures: List[float], dynatemp_mins: List[float],
...
+            dtype=dtype, 
+            pin_memory=pin_memory,
+        )
+        dry_allowed_lengths_t = torch.tensor(
+            dry_allowed_lengths,
+            device="cpu",
+            dtype=torch.int,
+            pin_memory=pin_memory,
+        )
+        dry_sequence_breakers_t = torch.tensor(
```
selalipop, line 768: Currently crashes at this line in some cases, I'm likely incorrectly treating sequence breaker array the same as the scalars above here

aphrodite/modeling/sampling_metadata.py
```diff
@@ -526,6 +538,10 @@ def from_sampling_metadata(
             xtc_thresholds += [params.xtc_threshold] * n_seqs
             xtc_probabilities += [params.xtc_probability] * n_seqs
             nsigmas += [params.nsigma] * n_seqs
+            dry_multipliers += [params.dry_multiplier] * n_seqs
+            dry_bases += [params.dry_base] * n_seqs
+            dry_allowed_lengths += [params.dry_allowed_length] * n_seqs
+            dry_sequence_breaker_ids += [params.dry_sequence_breaker_ids] * n_seqs
```
selalipop, line 544: This feels incorrect and the likely source of the crash, but it's not clear how to correctly handle dry_sequence_breaker_ids

selalipop:
This is probably in a pretty rough place with my lack of context on how certain pieces of this fit together, hence the draft

AlpinDale:
Thanks for the PR! I'll take a look very soon.

AlpinDale:
@selalipop I'm testing the PR and it seems to run fine, how do I reproduce the crash? I'm testing with the API server, using a curl command like this:

```console
curl http://localhost:2242/v1/completions \
-d '{
  "model": "NousResearch/Meta-Llama-3.1-8B-Instruct",
  "prompt": "Once upon a time",
  "dry_allowed_length": 2,
  "dry_multiplier": 1.12,
  "dry_base": 1.75,
  "dry_sequence_breakers": ["\\n", ":", "\"", "*"],
  "dry_penalty_last_n": 0
}' | jq .
```

AlpinDale:
Seems like the params are being passed correctly to the sampler, so there shouldn't be any issues where sequence breakers aren't passed to trigger the crash:

```diff
diff --git a/aphrodite/modeling/layers/sampler.py b/aphrodite/modeling/layers/sampler.py
index 4ac082df..5a605160 100644
--- a/aphrodite/modeling/layers/sampler.py
+++ b/aphrodite/modeling/layers/sampler.py
@@ -14,6 +14,7 @@ from aphrodite.common.sequence import (CompletionSequenceGroupOutput, Logprob,
                                        PromptLogprobs, SampleLogprobs,
                                        SamplerOutput, SequenceOutput)
 from aphrodite.triton_utils import HAS_TRITON
+from aphrodite.common.logger import log_once
 
 if HAS_TRITON:
     from aphrodite.modeling.layers.ops.sample import sample as sample_triton
@@ -200,6 +201,13 @@ class Sampler(nn.Module):
                 sampling_tensors.xtc_probabilities)
 
         if do_dry:
+            log_once(
+                level="INFO",
+                message="DRY has been enabled! Params: "
+                f"dry_multipliers={sampling_tensors.dry_multipliers}, "
+                f"dry_bases={sampling_tensors.dry_bases}, "
+                f"dry_allowed_lengths={sampling_tensors.dry_allowed_lengths}, "
+                f"dry_sequence_breaker_ids={sampling_tensors.dry_sequence_breaker_ids}")
             logits = _apply_dry(
                 logits,
                 sampling_tensors.prompt_tokens,
```

Logs:

```console
INFO:     Received request cmpl-fabe3eb67a0e4820a2df920e3e309dd9-0: prompt: , params: SamplingParams(custom_token_bans=None, dry_multiplier=1.12, dry_sequence_breaker_ids=[198, 25, 1, 9]), prompt_token_ids: [], lora_request: None, prompt_adapter_request: None.
INFO:     Added request cmpl-fabe3eb67a0e4820a2df920e3e309dd9-0.
INFO:     DRY has been enabled! Params: dry_multipliers=tensor([1.1172], device='cuda:0', dtype=torch.bfloat16), dry_bases=tensor([1.7500],  device='cuda:0', dtype=torch.bfloat16), dry_allowed_lengths=tensor([2], device='cuda:0', dtype=torch.int32),  dry_sequence_breaker_ids=tensor([[198,  25,   1,   9]], device='cuda:0')
```

aphrodite/common/sampling_params.py
```diff
@@ -199,7 +216,10 @@ class SamplingParams(
     xtc_threshold: float = 0.1
     xtc_probability: float = 0
     nsigma: float = 0.0
-
+    dry_multiplier: float = 0.0
+    dry_base: float = 1.75
+    dry_allowed_length: int = 2
+    dry_sequence_breaker_ids: List[float] = []
```
AlpinDale, line 222: Is this supposed to be a list of int?
```suggestion
    dry_sequence_breaker_ids: List[int] = []
```

AlpinDale:
Looked into this some more and it seems like this implementation isn't doing anything. This is because it has no access to prompt history or output tokens after they're generated. DRY will need to be implemented the same way as the other penalty samplers. A typical implementation won't work because, for efficiency's sake, we don't initialize prompt and output tokens tensors if penalties aren't being used. I'll fix this in a bit and have them initialize for DRY.

AlpinDale:
It seems to be working now.

selalipop:
The crash was due to an incorrectly shaped array when constructing `dry_sequence_breakers_t` in `sampling_metadata.py`

```
20T08:15:21.129660136Z ValueError: expected sequence of length 2 at dim 1 (got 3)
```

Trying to come up with a reproduction: could it be due to changing sequence breakers between requests that are batched together? (in other words, does seq_groups represent different requests?)

---

I'm also trying to "sanity check" with extremely high values and not seeing what I'd expect: the model seems to be coherent and happily repeat sequences, but I'd expect gibberish given the high penalties

Could I be completely off on how to use the tensors in `_apply_dry`? For example, in `_apply_dry` I'm calling `previous_token in sequence_breakers_ids`, is that a valid pattern given it's been turned into a tensor? 

I also see `__apply_min_tokens_penalty` is just using the non-tensor value from `sampling_metadata` to access stop_token_ids, would that be the more idiomatic approach here?

Sorry, my lack of familiarity makes me worried I might be making some flagrant mistakes 😬

selalipop:
Managed to reproduce the crash: I ran about 4 concurrent requests with DRY and without and eventually got a crash:

```ValueError: expected sequence of length 4 at dim 1 (got 0)```

At the instantiation of the `dry_sequence_breakers_t` IDs tensor

AlpinDale:
Thanks for that @discordianbelle, penalties actually work now. 

I'll add a test to run this batched and see if the issue still persists.

AlpinDale:
We're good here for now. Thanks for the help everyone.

I'll merge this then add range (along with kcpp improvements) in a follow-up PR.

PR #837, opened 2024-11-26
feat: add sampler_priorty

AlpinDale:
Since #259 was never finished, this is a re-implementation. Entirely different.

The corresponding integers for each sampler is arbitrary at the moment, will change once I check out how other projects do it.


LostRuins:
LGTM

Issue #836, opened 2024-11-25
[Feature]: pass-through parameter from request to model.forward (already implemented)


qpwo:
### 🚀 The feature, motivation and pitch

It took me ages to figure out how to pass a custom parameter from the API through to the model.forward call. This is very useful for running little experiments and making custom logs.

Should I dress it up in a PR?

It would be disabled by default. I guess you'd do `aphrodite run --enable-passthrough-param` or similar.

## Example

```py
## example request:
client.completions.create(
    model='70b',
    prompt='I was',
    extra_body={"passthrough": {"save_activations": True}},
)

## example modification of LlamaForCausalLM.forward:
def forward(
    self,
    input_ids: torch.Tensor,
    positions: torch.Tensor,
    kv_caches: List[torch.Tensor],
    attn_metadata: AttentionMetadata,
    intermediate_tensors: Optional[IntermediateTensors] = None,
    passthrough: Optional[Dict] = None,
) -> Union[torch.Tensor, IntermediateTensors]:
    passthrough = passthrough or {}
    save_activations = bool(passthrough.get("save_activations", False)) # example 1
    scale_gate_up = float(passthrough.get('scale_gate_up', 1.0)) # example 2
    model_output = self.model(
        input_ids, positions, kv_caches, attn_metadata, intermediate_tensors,
        save_activations=save_activations,
        scale_gate_up=scale_gate_up,
    )
    return model_output
```

### Alternatives

I could not find any alternatives but would be quite happy to learn of any.

### Additional context

PR is mostly written already, I would just tidy it if you want this feature.

AlpinDale:
Doesn't seem particularly difficult to maintain, so go ahead!

PR #838, opened 2024-11-26
passthrough arg from api to model.forward

qpwo:
Please feel welcome to make any and all changes without consulting me!

Addresses #836

About this PR (and my limited understanding of how aphrodite works)

- A `passthrough` will come in from the API and get to either SamplingParams or PoolingParams.
- The goal is to pass it around and get it all the way to `model_executable(**execute_model_kwargs)`.
- So we need to get `passthrough` onto `ModelInputForXPU`, the open vino thing, `ModelInputForGPU`, `ModelInputForNeuron`, and `ModelInputForTPU`.
- I only implemented distributed stuff for GPU I think.
- A function has access to the `passthrough` if it gets SamplingParams or PoolingParams or LLMInputs or EncoderDecoderLLMInputs.
   -   I tried to keep redundant passing of passthrough to a minimum
- I think `passthrough' travels between GPUs and nodes/machines along ModelInputForGPUBuilder.InterDataForSeqGroup?
- Not sure this is how you're supposed to do all this.
- Seems to work for me.

There are a couple little todos that might need done, i cant tell

### DIFF
--- a/aphrodite/common/passthrough.py
+++ b/aphrodite/common/passthrough.py
@@ -1,0 +1,20 @@
+from typing import Optional, Union
+from typing_extensions import TypedDict
+
+
+class Passthrough(TypedDict):
+    """
+    (development only) arguments passed through to the model.forward call from the API
+    """
+    foo: int
+    bar: str
+
+def try_get_passthrough(params: Optional[Union["SamplingParams", "PoolingParams"]]) -> Optional[Passthrough]:
+    passthrough = None
+    if isinstance(params, SamplingParams):
+        passthrough = params.passthrough
+    elif isinstance(params, PoolingParams) and isinstance(params.additional_data, dict):
+        passthrough = params.additional_data.get("passthrough")
+    return passthrough
+from aphrodite.common.pooling_params import PoolingParams
+from aphrodite.common.sampling_params import SamplingParams
--- a/aphrodite/common/passthrough.py
+++ b/aphrodite/common/passthrough.py
@@ -2,7 +2,6 @@ from typing import Optional, Union
 
 from typing_extensions import TypedDict
 
-
 class Passthrough(TypedDict):
     """
     (development only) arguments passed through to the model.forward call
@@ -22,5 +21,6 @@ def try_get_passthrough(
         passthrough = params.additional_data.get("passthrough")
     return passthrough
 
+# prevent circular import:
 from aphrodite.common.pooling_params import PoolingParams  # noqa: E402
 from aphrodite.common.sampling_params import SamplingParams  # noqa: E402
--- a/aphrodite/common/passthrough.py
+++ b/aphrodite/common/passthrough.py
@@ -1,20 +1,26 @@
 from typing import Optional, Union
+
 from typing_extensions import TypedDict
 
 
 class Passthrough(TypedDict):
     """
-    (development only) arguments passed through to the model.forward call from the API
+    (development only) arguments passed through to the model.forward call
+    from the API
     """
     foo: int
     bar: str
 
-def try_get_passthrough(params: Optional[Union["SamplingParams", "PoolingParams"]]) -> Optional[Passthrough]:
+def try_get_passthrough(
+    params: Optional[Union["SamplingParams", "PoolingParams"]]
+)-> Optional[Passthrough]:
     passthrough = None
     if isinstance(params, SamplingParams):
         passthrough = params.passthrough
-    elif isinstance(params, PoolingParams) and isinstance(params.additional_data, dict):
+    elif isinstance(params, PoolingParams) and \
+        isinstance(params.additional_data, dict):
         passthrough = params.additional_data.get("passthrough")
     return passthrough
-from aphrodite.common.pooling_params import PoolingParams
-from aphrodite.common.sampling_params import SamplingParams
+
+from aphrodite.common.pooling_params import PoolingParams  # noqa: E402
+from aphrodite.common.sampling_params import SamplingParams  # noqa: E402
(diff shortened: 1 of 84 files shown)


qpwo:
Here's a (rather long) video of it working

https://github.com/user-attachments/assets/8b601fb1-12bd-4789-8a57-ccdbd827c0ff

(Again, if I drop off, feel free to close it or rewrite it or whatever. No need to ask me for permission.)

qpwo:
I should have time for one or two rounds of improvements though

qpwo:
Lmk what I should do with this

qpwo:
just rebased

aphrodite/common/passthrough.py
```diff
@@ -0,0 +1,26 @@
...
+    passthrough = None
+    if isinstance(params, SamplingParams):
+        passthrough = params.passthrough
+    elif isinstance(params, PoolingParams) and \
+        isinstance(params.additional_data, dict):
+        passthrough = params.additional_data.get("passthrough")
+    return passthrough
+
+from aphrodite.common.pooling_params import PoolingParams  # noqa: E402
+from aphrodite.common.sampling_params import SamplingParams  # noqa: E402
```
AlpinDale, line 26: Why are these imported at the bottom?

aphrodite/endpoints/llm.py
```diff
@@ -92,6 +92,8 @@ class LLM:
         serving, use the :class:`~aphrodite.AsyncAphrodite` class instead.
     """
 
+    # TODO:Luke do we need `enable_passthrough_param` stuff in llm.py?
```
AlpinDale, line 95: You can leave it for now, since passthrough seems to be a REST-API first-class.

aphrodite/modeling/models/llama.py
```diff
@@ -340,6 +341,11 @@ def forward(
         hidden_states, _ = self.norm(hidden_states, residual)
         return hidden_states
 
+_printed = set()
+def print_once(key, *args, **kwargs):
```
AlpinDale, line 345: We have a `log_once()` function in `aphrodite.common.logger`. You can use that instead of this.

AlpinDale (commented):
Left a few comments, mostly just nits for now.

This PR doesn't seem to be complete yet, so I can't make a decisive approval. Can you commit an example usage script for this? I'm still struggling to see how it can be used in its current state 😅

50h100a:
Setting aside for a moment ways the PR could be improved, I'm also struggling to find a motivating use-case for this as it is now.

Do you have a specific example of something this would allow you to do, that would be much harder without it?

aphrodite/common/passthrough.py
qpwo: prevent circular import

qpwo:
> Do you have a specific example of something this would allow you to do, that would be much harder without it?

I am adding steering vectors to llama 405b. So I can just put `save_activation=True` or `steer_with=1234` in `passthrough`. If PR is not a good fit then I'll close it and just use my fork.

AlpinDale:
If you need Steering/Control Vectors in particular, I've been working on this PR to enable support for it: #604

qpwo:
Oh sweet! But yeah I wanted something a bit more flexible. Quickly make changes in the `<model>.py`code itself. Hmm. So should I finish this PR or not?

qpwo:
I also was adding custom parameters to the samplers. It's nice to not have to change code in 20 places to try something out. But this is your repo not mine! Lmk

AlpinDale:
If you think you can make the changes as unintrusive as possible, I'm not against it. What do you think @50h100a ?

In case we do merge the PR, I'd appreciate it if you helped with the maintenance related to passthrough when needed. Since we don't have many maintainers, we try and add only the features we think we can comfortably maintain.

qpwo:
Would be happy to try and patch it up when issues arise. I must admit I am still a bit confused about how the scheduling, queuing, inter-node comms, etc still work. Do you have a doc  anywhere explaining like model_runner.py and scheduler.py?

qpwo:
Everything that takes a `passthrough` has `passthrough: Optional[Passthrough] = None`, so I think this is a relatively safe PR to merge as-is, despite the big diff. (Biggest thing is I haven't tested non-cuda arches.) I have been using it the last few days and I've found it pretty handy.

My latest pattern is to have like `passthrough.action: Save | Steer` then in `layer.forward` I can do `if action.type == 'steer': ...` and it's a bit more ergonomic than using hooks or wrapper classes for everything. But it's really only practical for running little experiments.

I won't mind at all if you close it -- your project! I am not attached to this

qpwo:
Oh by the way, the reason I am using aphrodite for my experiments is that it runs twice as fast as the leanest meanest llama inference code I could come up with! I realize that quick experiments are not your main target use case. But thanks a lot for all the cuda kernels

AlpinDale:
> Would be happy to try and patch it up when issues arise. I must admit I am still a bit confused about how the scheduling, queuing, inter-node comms, etc still work. Do you have a doc anywhere explaining like model_runner.py and scheduler.py?

There's unfortunately no explicit documentation for the internal sequence processing code, I'll get to it as soon as I can. For now, you'd have to reply on the docstrings and comments, which I think should be descriptive enough for most cases.

50h100a:
This PR can be radically simplified by using kwargs unpacking in the `model_executable` calls in the model runners - no need to make changes to every single model implementation. Just use `**passthrough`, treating it as a possibly-empty dict of extra args.
As it's a debug feature, failing ungracefully when unsupported kwargs are passed to a model seems acceptable.

I suspect the things upstream can be simplified as well, and certainly generalized into an anonymous dict, but I'm on less familiar ground there.

qpwo:
Yeah I would've done that if I knew how. I don't really understand the code so I was just mimicking the style of the other stuff passed explicitly. If you did the kwargs thing then the diff might improve my understanding

qpwo:
I am confused also about why most stuff gets explicitly tensorified/detensorified and the code still ran when i just didn't do that at all. Maybe it relates to torch.compile or XPUs or something?

PR #849, opened 2024-11-28
DRY Fix: Add output_tokens to sampler

selalipop:
DRY Sampler currently only looks at prompt_tokens


AlpinDale:
@discordianbelle can you please check if this branch works for you?

discordianbelle:
Looks like this is done already, but I can also confirm this is now working as expected, and also no longer needs the very high hyperparameters values from my fork. Thanks!

PR #856, opened 2024-12-02
sampler: optimize DRY performance using z-algorithm

AlpinDale:
Reference: oobabooga/text-generation-webui#6047

Needs manual testing, but the test units pass.

### DIFF
--- a/aphrodite/modeling/layers/sampler.py
+++ b/aphrodite/modeling/layers/sampler.py
@@ -674,10 +674,11 @@ def _apply_dry(
         if multiplier == 0:
             continue
 
+        seq_breakers = set(sequence_breakers_ids[i].tolist())
         input_ids_list = input_ids_row.tolist()
         last_token = input_ids_list[-1]
 
-        if last_token in sequence_breakers_ids:
+        if last_token in seq_breakers:
             continue
 
         range_limit = ranges[i].item()
--- a/aphrodite/modeling/layers/sampler.py
+++ b/aphrodite/modeling/layers/sampler.py
@@ -628,7 +628,7 @@ def _apply_dry(
 ) -> torch.Tensor:
     """
     Apply Don't Repeat Yourself (DRY) sampling to the logits.
-    
+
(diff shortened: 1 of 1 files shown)


aphrodite/modeling/layers/sampler.py
```diff
@@ -638,69 +638,85 @@ def _apply_dry(
...
-        multiplier = multipliers[i]  
         base = bases[i]
 
-        for token, match_length in match_lengths.items():
-            if match_length >= allowed_length and token < vocab_size:
+        for idx, match_length in enumerate(z_array[:-1]):
+            if match_length >= allowed_length:
+                next_token = input_ids_list[idx + 1]
+                if (next_token >= vocab_size or next_token in
+                    sequence_breakers_ids):
```
mehdieverai, line 711: seq_breakers, no?

gitzaidi, line 711: seq_breakers instead of sequence_breakers_ids I think

aphrodite/modeling/layers/sampler.py
```diff
@@ -638,69 +638,85 @@ def _apply_dry(
...
-                match_lengths[next_token] = match_length
-
+            search_start = max(0, len(input_ids_list) - range_limit)
+
+        # Find max match length based on sequence breakers
+        max_match_length = 0
+        MAX_LENGTH = min(len(input_ids_list), 1000)  # Prevent overflow
+        while (max_match_length < MAX_LENGTH and 
+               input_ids_list[len(input_ids_list) - max_match_length - 1] 
+               not in sequence_breakers_ids):
```
gitzaidi, line 695: same, seq_breakers instead of sequence_breakers_ids I think

gitzaidi:
Sure, [here](https://github.com/PygmalionAI/aphrodite-engine/pull/857)

p-e-w:
When testing the original implementation, it was found that limiting match length (which turns the quadratic algorithm into a linear one because the inner loop has bounded iteration count) gives the same performance as the substantially more complicated Z-Algorithm. Was this not true in your tests?

PR #868, opened 2024-12-04
Rewrite DRY sampler to be a lot faster

50h100a:
Upgraded from extremely slow to pretty slow.

Possible issues:
 - Calculating the padding around ln 642 seems dumb, surely we have that somewhere?
 - Suspect the ln 681 loop is where the slowness is coming from now, but I'm not sure. Also not sure how to tensorize it atm.


Nero10578:
This is cooking. Thanks for the fix.

PR #888, opened 2024-12-14
Add Repetition Range ('rep_range')

discordianbelle:
Adds a range to the repetition penalties (all samplers under do_penalties)

Counts back from the current token, applying to all output tokens within range, then prompt tokens if the range extends that far

The most expensive operations are just simple slicing operations which are relatively fast in PyTorch.

### DIFF
--- a/aphrodite/common/sampling_params.py
+++ b/aphrodite/common/sampling_params.py
@@ -229,6 +229,7 @@ class SamplingParams(
     presence_penalty: float = 0.0
     frequency_penalty: float = 0.0
     repetition_penalty: float = 1.0
+    rep_range: Optional[int] = None
     no_repeat_ngram_size: int = 0
     temperature: float = 1.0
     dynatemp_min: float = 0.0
@@ -287,6 +288,7 @@ class SamplingParams(
         "presence_penalty": 0.0,
         "frequency_penalty": 0.0,
         "repetition_penalty": 1.0,
+        "rep_range": None,
         "no_repeat_ngram_size": 0,
         "temperature": 1.0,
         "dynatemp_min": 0.0,
@@ -400,6 +402,9 @@ class SamplingParams(
         if self.repetition_penalty < 1.0:
             raise ValueError("repetition_penalty must be in [1, inf), got "
                              f"{self.repetition_penalty}.")
+        if self.rep_range is not None and self.rep_range < 1:
+            raise ValueError("rep_range must be at least 1 if specified, got "
+                             f"{self.rep_range}.")
         if self.temperature < 0.0:
             raise ValueError(
                 f"temperature must be non-negative, got {self.temperature}.")
--- a/aphrodite/common/sampling_params.py
+++ b/aphrodite/common/sampling_params.py
@@ -9,12 +9,12 @@ import torch
 from loguru import logger
 from typing_extensions import Annotated
 
-from aphrodite import envs
+import os
 
 _SAMPLING_EPS = 1e-5
 _MAX_TEMP = 1e-2
 
-APHRODITE_NO_DEPRECATION_WARNING = envs.APHRODITE_NO_DEPRECATION_WARNING
+APHRODITE_NO_DEPRECATION_WARNING = bool(int(os.environ.get("APHRODITE_NO_DEPRECATION_WARNING", "0")))
 
 
 class SamplingType(IntEnum):
--- a/aphrodite/modeling/layers/sampler.py
+++ b/aphrodite/modeling/layers/sampler.py
@@ -10,7 +10,7 @@ import torch.nn as nn
 from loguru import logger
 
 import aphrodite._custom_ops as ops
-from aphrodite import envs
+import os
 from aphrodite.common.sampling_params import SamplingType
 from aphrodite.common.sequence import (CompletionSequenceGroupOutput, Logprob,
                                        PromptLogprobs, SampleLogprobs,
@@ -34,7 +34,7 @@ _TEMPERATURE_MINIMUM = 2e-5
 
 # If enabled, we switch to a more performant implementation
 # of top-k and top-p
-APHRODITE_USE_SAMPLING_KERNELS = envs.APHRODITE_USE_SAMPLING_KERNELS
+APHRODITE_USE_SAMPLING_KERNELS = bool(int(os.environ.get("APHRODITE_USE_SAMPLING_KERNELS", "0")))
 
 
 class SamplerID(IntEnum):
--- a/aphrodite/modeling/layers/sampler.py
+++ b/aphrodite/modeling/layers/sampler.py
@@ -272,7 +272,7 @@ class Sampler(nn.Module):
                     sampling_tensors.output_tokens,
                     sampling_tensors.presence_penalties,
                     sampling_tensors.frequency_penalties,
-                    sampling_tensors.repetition_penalties.
+                    sampling_tensors.repetition_penalties,
                     rep_range=rep_range)
 
             elif sampler_id == SamplerID.NO_REPEAT_NGRAM and \
--- a/aphrodite/modeling/layers/sampler.py
+++ b/aphrodite/modeling/layers/sampler.py
@@ -272,7 +272,8 @@ class Sampler(nn.Module):
                     sampling_tensors.output_tokens,
                     sampling_tensors.presence_penalties,
                     sampling_tensors.frequency_penalties,
-                    sampling_tensors.repetition_penalties)
+                    sampling_tensors.repetition_penalties.
+                    rep_range=rep_range)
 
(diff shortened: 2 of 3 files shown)


aphrodite/common/sampling_params.py
```diff
@@ -9,12 +9,12 @@
...
 from typing_extensions import Annotated
 
-from aphrodite import envs
+import os
 
 _SAMPLING_EPS = 1e-5
 _MAX_TEMP = 1e-2
 
-APHRODITE_NO_DEPRECATION_WARNING = envs.APHRODITE_NO_DEPRECATION_WARNING
+APHRODITE_NO_DEPRECATION_WARNING = bool(int(os.environ.get("APHRODITE_NO_DEPRECATION_WARNING", "0")))
```
AlpinDale, line 17: Why this change?

aphrodite/common/sampling_params.py
```diff
@@ -400,6 +402,9 @@ def _verify_args(self) -> None:
         if self.repetition_penalty < 1.0:
             raise ValueError("repetition_penalty must be in [1, inf), got "
                              f"{self.repetition_penalty}.")
+        if self.rep_range is not None and self.rep_range < 1:
```
AlpinDale, line 405: We should probably allow 0 for infinite (or all tokens) range. Unless other inference software do it this way.

aphrodite/modeling/layers/sampler.py
```diff
@@ -34,7 +34,7 @@
 
 # If enabled, we switch to a more performant implementation
 # of top-k and top-p
-APHRODITE_USE_SAMPLING_KERNELS = envs.APHRODITE_USE_SAMPLING_KERNELS
+APHRODITE_USE_SAMPLING_KERNELS = bool(int(os.environ.get("APHRODITE_USE_SAMPLING_KERNELS", "0")))
```
AlpinDale, line 37: Same as before.

aphrodite/modeling/layers/sampler.py
```diff
@@ -507,25 +508,39 @@ def _apply_penalties(logits: torch.Tensor, prompt_tokens_tensor: torch.Tensor,
...
+    
+    _, prompt_mask = _get_bin_counts_and_mask(prompt_tokens_tensor, vocab_size, num_seqs)
+    output_bin_counts, output_mask = _get_bin_counts_and_mask(output_tokens_tensor, vocab_size, num_seqs)
 
     repetition_penalties = repetition_penalties[:, None].repeat(1, vocab_size)
     repetition_penalties[~(prompt_mask | output_mask)] = 1.0
     logits = torch.where(logits > 0, logits / repetition_penalties,
                          logits * repetition_penalties)
 
-    # We follow the definition in OpenAI API.
```
AlpinDale, line 522: Don't remove comment.

aphrodite/modeling/layers/sampler.py
```diff
@@ -272,7 +272,8 @@ def forward(
                     sampling_tensors.output_tokens,
                     sampling_tensors.presence_penalties,
                     sampling_tensors.frequency_penalties,
-                    sampling_tensors.repetition_penalties)
+                    sampling_tensors.repetition_penalties,
+                    rep_range=rep_range)
```
AlpinDale, line 276:
```suggestion
                    sampling_tensors.rep_range)
```

This parameter needs to be added to the sampling_metadata module.

aphrodite/modeling/layers/sampler.py
```diff
@@ -507,25 +508,39 @@ def _apply_penalties(logits: torch.Tensor, prompt_tokens_tensor: torch.Tensor,
                      output_tokens_tensor: torch.Tensor,
                      presence_penalties: torch.Tensor,
                      frequency_penalties: torch.Tensor,
-                     repetition_penalties: torch.Tensor) -> torch.Tensor:
+                     repetition_penalties: torch.Tensor,
+                     rep_range: Optional[int] = None) -> torch.Tensor:
```
AlpinDale, line 512:
```suggestion
                     rep_range: torch.Tensor) -> torch.Tensor:
```

Must be tensorized after it's added to sampling_metadata. If we treat it a single integer, it'll apply to all sequences within the batch; tensorizing this will allow us to match the batch dimension, in case other sequences may want different ranges.

aphrodite/modeling/layers/sampler.py
```diff
@@ -507,25 +508,39 @@ def _apply_penalties(logits: torch.Tensor, prompt_tokens_tensor: torch.Tensor,
...
     num_seqs, vocab_size = logits.shape
-    _, prompt_mask = _get_bin_counts_and_mask(prompt_tokens_tensor, vocab_size,
-                                              num_seqs)
-    output_bin_counts, output_mask = _get_bin_counts_and_mask(
-        output_tokens_tensor, vocab_size, num_seqs)
+    
+    if rep_range is not None and rep_range > 0:
+        # Just take the last rep_range tokens from output_tokens_tensor
+        # This is much more efficient as we're only looking at recent history
+        output_tokens_tensor = output_tokens_tensor[:, -rep_range:]
```
AlpinDale, line 519: This is applying the same range to all sequences in batch, no?

Also creating a new tensor here is probably less efficient. I think the slicing should be done in the bin counting function above this.

aphrodite/modeling/layers/sampler.py
```diff
@@ -507,25 +508,39 @@ def _apply_penalties(logits: torch.Tensor, prompt_tokens_tensor: torch.Tensor,
...
+        
+        # Only include prompt tokens if we're still within rep_range of them
+        prompt_end_idx = prompt_tokens_tensor.size(1)
+        output_len = output_tokens_tensor.size(1)
+        if output_len < rep_range:
+            # Calculate how many prompt tokens we should include
+            prompt_tokens_to_include = min(rep_range - output_len, prompt_end_idx)
+            prompt_tokens_tensor = prompt_tokens_tensor[:, -prompt_tokens_to_include:]
```
AlpinDale, line 527: Not all sequences in a batch have the same output len (and consequently may not include the same number of prompt tokens).

aphrodite/modeling/layers/sampler.py
```diff
@@ -507,25 +508,39 @@ def _apply_penalties(logits: torch.Tensor, prompt_tokens_tensor: torch.Tensor,
...
+        else:
+            # If we have enough output tokens, ignore prompt completely
+            prompt_tokens_tensor = torch.empty((num_seqs, 0), dtype=torch.long, device=logits.device)
```
AlpinDale, line 530: Do we need to do this here? I think most of the range ops can be done in bin counting.

AlpinDale (changes requested):
Thanks for the PR. There's a few issues in its current state, as outlined in my comments. Mainly, this seems designed for sequential inference, and will fail for batched inference, due to no handling for per-sequence differences within a batch.

Also please run `./formatting.sh` to fix linting issues.

aphrodite/modeling/layers/sampler.py
discordianbelle: I'll refactor it, thanks again

aphrodite/modeling/layers/sampler.py
discordianbelle: Agreed, I'll have to go over it again with batching in mind

aphrodite/modeling/layers/sampler.py
discordianbelle: Agreed, I'll have to go over it again with batching in mind, and also agreed about rolling in the slicing

aphrodite/modeling/layers/sampler.py
discordianbelle: Great point, will handle

aphrodite/modeling/layers/sampler.py
discordianbelle: I'll fix on Monday!

aphrodite/modeling/layers/sampler.py
discordianbelle: o7
Sorry, danger of AI-assisted programming

aphrodite/modeling/layers/sampler.py
discordianbelle: See other instance

aphrodite/common/sampling_params.py
discordianbelle: You're absolutely right, I'll implement on Monday

aphrodite/common/sampling_params.py
discordianbelle: After installing as editible and then modifying the files, I was getting circular import errors. As far as I could tell it was because envs.py is in ./aphrodite/ and not ./aphrodite/common/

I was trying to make the least impactful change that still let it run, so I didn't want to move envs.py

PR #943, opened 2024-12-21
Fix SentencePieceTokenizer error when generating on Mistral Large 2411 with `--tokenizer-mode mistral`

khanonnie:
When starting a generation using Mistral Large 2411 (or 2407) with `--tokenizer-mode mistral` set, the generation fails and an error is thrown: `TypeError: SentencePieceTokenizer.encode() missing 2 required positional arguments: 'bos' and 'eos'`. The stack trace points to the guided decoding processor.

Tweaking `aphroditeLogitsProcessor#build_aphrodite_token_enforcer_tokenizer_data` to match the latest version of [lm-format-enforcer's vllm implementation](https://github.com/noamgat/lm-format-enforcer/blob/main/lmformatenforcer/integrations/vllm.py#L35-L46), with a MistralTokenizer case and passing `vocab_size` into `build_token_enforcer_tokenizer_data`, appears to fix this issue.

### DIFF
--- a/aphrodite/modeling/guided_decoding/lm_format_enforcer_logits_processors.py
+++ b/aphrodite/modeling/guided_decoding/lm_format_enforcer_logits_processors.py
@@ -10,6 +10,7 @@ from lmformatenforcer.integrations.transformers import (
 from transformers import PreTrainedTokenizerBase
 
 import aphrodite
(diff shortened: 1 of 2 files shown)


AlpinDale:
Thanks for this. The current fix was to switch to outlines when using mistral tokenizer, but this is much better.

PR #1259, opened 2025-03-11
feat: migrate docs to starlight

ahme-dev:


docs/CNAME
AlpinDale, line 1: Any reason why this is removed?

ahme-dev: Will be adding the deploy action and re-adding that

PR #1283, opened 2025-03-24
docs: fix missing parameter in rocm installation guide

dsk7:
The command two lines above has the missing parameter -t to apply a tag, this line does not.
This PR adds the missing parameter.


AlpinDale:
Thanks for the fix!

PR #1293, opened 2025-04-14
scheduler changes to allow LoRA + Chunked Prefill

Nero10578:
Seems to be working fine with this change which are the same as what allowed VLLM to enable chunked prefill along with LoRA. https://github.com/vllm-project/vllm/pull/9057/files


AlpinDale (approved):
Thanks!

PR #1626, opened 2026-04-16
[attention] fix: initialize is_default_backend_supported for generic fallback

Th-Underscore:


AlpinDale:
Thanks for the PR; however this is no longer required as of v0.20.0.

PR #1717, opened 2026-07-23
Support for logit bias in specdec

50h100a:
Support for logit bias in specdec, tests for it, and a NotImplementedError to scream if a LogitsProcessor is used when it shouldn't be.


AlpinDale:
This breaks `tests/v1/logits_processors/test_custom_offline.py::test_rejects_custom_logitsprocs`

PR #1766, opened 2026-08-13
fix: thinking_token_budget has no effect when async scheduling is enabled

Nakanokensetsu:
## Summary

`ThinkingBudgetStateHolder` tracks each request's `<think>`/`</think>` state and forces the reasoning end tokens onto the logits once `thinking_token_budget` is exceeded, via two methods. Both had a wiring gap:

- `update_state()` (advances the per-request think/end state machine from the latest sampled tokens) was only ever called from `sync_batch()` (batch add/remove/move bookkeeping), never once per decode step. A request's think state was never advanced past the moment it entered the batch, so budget overrun was never detected once generation was under way.
- `apply_to_logits()` (forces the reasoning end token(s) into the logits) was only wired into `rejection_sampler.py`, the speculative-decoding sampler path. `Sampler.forward()`, used whenever speculative decoding is not active, never called it, so even a correctly-tracked budget overrun was never forced onto the logits.

Together, `thinking_token_budget` silently had no effect for any request on the normal (non-speculative-decoding) sampling path once async scheduling is enabled — which is the default for compatible executors (`AphroditeConfig.__post_init__`, `scheduler_config.async_scheduling is None -> True`). Reasoning ran unbounded regardless of the configured budget, up to `max_tokens`.

The existing tests in `test_thinking_token_budget.py` didn't catch this because their server fixtures explicitly pass `--no-async-scheduling`. Under that config, `_make_sampling_metadata()` happens to call `update_state()` from a different, sufficient path, masking the bug.

## Reproduction

Manual repro against `Qwen/Qwen3-0.6B` with async scheduling left at its default (enabled), `thinking_token_budget=5`, `max_tokens=100`:

```
before: reasoning_token_count=None, total_decode_tokens=100
        (budget ignored entirely, ran to max_tokens without ever closing </think>)
after:  reasoning_token_count=5, total_decode_tokens=17
        (budget respected exactly, natural completion)
```

## Fix

- Call `update_state()` every decode step from `GPUModelRunner._sample()`.
- Call `apply_to_logits()` from `Sampler.forward()`, mirroring the existing `rejection_sampler.py` call.

Both paths are now covered regardless of speculative decoding or async scheduling state.

## Test plan

- [x] Existing `test_thinking_token_budget.py` suite (6 non-MTP cases, `default`/`auto_config` params) passes unchanged
- [x] Added a new `async_scheduling` server fixture/param to `test_thinking_token_budget_limits_reasoning` that leaves async scheduling at its default instead of disabling it, so this regression is covered going forward
- [x] Manual before/after repro documented above

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/aphrodite/v1/worker/gpu_model_runner.py
+++ b/aphrodite/v1/worker/gpu_model_runner.py
@@ -3357,10 +3357,7 @@ class GPUModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin, ECConnec
         # way. Call it here every step, before sampling, using this step's
         # freshly-updated token lists.
         thinking_budget_state_holder = sampling_metadata.thinking_budget_state_holder
-        if (
-            thinking_budget_state_holder is not None
-            and thinking_budget_state_holder.has_tracked_requests()
-        ):
+        if thinking_budget_state_holder is not None and thinking_budget_state_holder.has_tracked_requests():
             thinking_budget_state_holder.update_state(
                 sampling_metadata.output_token_ids,
                 sampling_metadata.spec_token_ids,
--- a/aphrodite/v1/worker/gpu_model_runner.py
+++ b/aphrodite/v1/worker/gpu_model_runner.py
@@ -3349,6 +3349,22 @@ class GPUModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin, ECConnec
         # Update output token ids with tokens sampled in last step
         # if async scheduling and required by current sampling params.
         self.input_batch.update_async_output_token_ids()
+        # ThinkingBudgetStateHolder.update_state() drives the per-request
+        # think/end state machine forward; it was previously never called
+        # from the decode loop (only sync_batch() ran, on batch add/remove/
+        # move), so budget overrun was never detected past the first step
+        # and thinking_token_budget had no effect once generation was under
+        # way. Call it here every step, before sampling, using this step's
+        # freshly-updated token lists.
+        thinking_budget_state_holder = sampling_metadata.thinking_budget_state_holder
+        if (
+            thinking_budget_state_holder is not None
+            and thinking_budget_state_holder.has_tracked_requests()
+        ):
+            thinking_budget_state_holder.update_state(
+                sampling_metadata.output_token_ids,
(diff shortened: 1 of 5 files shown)


tests/entrypoints/openai/chat_completion/test_thinking_token_budget.py
```diff
@@ -142,11 +142,36 @@ def server_qwen35_fp8_mtp_tp2():
...
+    # thinking_token_budget is not yet supported by the V2 model runner.
+    env_dict = {"APHRODITE_USE_V2_MODEL_RUNNER": "0"}
+    with RemoteOpenAIServer(MODEL_NAME, args, env_dict=env_dict) as remote_server:
+        yield remote_server
+
+
 @pytest_asyncio.fixture
-async def client(request, server, server_with_auto_reasoning_config):
+async def client(request, server, server_with_auto_reasoning_config, server_async_scheduling):
     server_map = {
```
AlpinDale, line 171: We're now launching 3 servers with each occupying 40% of the available VRAM.

aphrodite/v1/worker/gpu_model_runner.py
```diff
@@ -3413,6 +3349,19 @@ def _sample(
...
+        # ThinkingBudgetStateHolder.update_state() drives the per-request
+        # think/end state machine forward; it was previously never called
+        # from the decode loop (only sync_batch() ran, on batch add/remove/
+        # move), so budget overrun was never detected past the first step
+        # and thinking_token_budget had no effect once generation was under
+        # way. Call it here every step, before sampling, using this step's
+        # freshly-updated token lists.
+        thinking_budget_state_holder = sampling_metadata.thinking_budget_state_holder
+        if thinking_budget_state_holder is not None and thinking_budget_state_holder.has_tracked_requests():
+            thinking_budget_state_holder.update_state(
```
AlpinDale, line 3361: This seems to update spec state before draft tokens get refreshed? Since the `update_state()` call is preceding `update_async_spec_token_ids()`.

AlpinDale (changes requested):
Thanks and sorry for not looking at this sooner. Please fix the two comments. It'd be nice if we could have this fix the for V1 runner too, but having it for V2 is good enough.

PR #1769, opened 2026-08-30
feat: resolve oci:// model references via llmman serve

ericcurtin:
## Summary

Adds an `oci://` scheme so a model published as a [CNCF ModelPack](https://github.com/modelpack/model-spec) OCI artifact can be used anywhere a HuggingFace repo id can:

```bash
aphrodite run oci://ghcr.io/org/model:tag
```

Model distribution is increasingly moving to OCI registries -- the same registries, credentials, mirroring and air-gap tooling a deployment already uses for container images.

## How it works

`ModelConfig.maybe_pull_model_tokenizer_for_runai` is already the one rewrite hook that runs before anything else touches `model=`/`tokenizer=`, so the `oci://` branch goes there. The two schemes are disjoint, so it dispatches and returns early -- the object-storage path is reached only when no `oci://` reference is present, and is otherwise bit-for-bit unchanged.

Acquisition is delegated to a running [`llmman serve`](https://github.com/llmmanorg/llmman) rather than hand-rolled: llmman already implements the ModelPack media types, registry auth, resumable blob download and a content-addressed store.

New `aphrodite/transformers_utils/llmman.py` is the client, stdlib-only (`urllib`), **no new dependency**:

- `GET /api/version` probes reachability *and* identity; a server answering without a `version` field is reported as "not an llmman daemon", worth distinguishing from nothing listening.
- `POST /api/pull` streams NDJSON so a multi-gigabyte fetch is not silent. An error arrives **in-band at HTTP 200**, and a stream that ends without `success` is also a failure -- both are errors, not a completed pull.
- `llmman resolve --no-pull` reports where the bytes landed. The daemon deliberately exposes no local path, so the CLI is the documented interface; `--no-pull` guarantees it only reports on what `/api/pull` already fetched, keeping the daemon the only thing that touches the network.
- `LLMMAN_HOST` is honoured with llmman's own parsing, including rewriting a wildcard bind (`0.0.0.0`, `[::]`) to loopback.

`aphrodite/transformers_utils/oci_utils.py` handles the scheme, deliberately mirroring the shape of the neighbouring `runai_utils.py` (including the `str()` cast so `pathlib.Path` inputs work).

A pull needs **both** the daemon reachable and the binary on `PATH` (or `APHRODITE_LLMMAN_BIN`); each missing piece has its own actionable error. Neither is required unless an `oci://` reference is actually used.

## Design notes

- **Explicit scheme, no sniffing.** A bare `registry/name:tag` is indistinguishable from a HuggingFace repo id (`org/model`); guessing would silently hijack existing `--model org/model` deployments.
- **`model_weights` keeps the reference the user typed**, matching what the runai path does with its URL.
- **One pull for a shared reference.** A `--tokenizer` naming the same image reuses the resolved directory -- a ModelPack image is pulled whole, so the tokenizer is already there.

## Testing

Two new files under `tests/transformers_utils/`. `test_llmman.py` (8 cases) runs against a **real HTTP server on a loopback port**, not mocks, so the NDJSON streaming contract is genuinely exercised: `/api/version` accepted, a non-llmman server rejected, nothing-listening reported actionably; pull success with forwarded byte progress and the exact request body asserted; in-band error at HTTP 200; a stream ending without `success`; non-OK status; a non-JSON diagnostic tolerated.

`test_oci_utils.py`: scheme detection incl. case-insensitivity and `pathlib.Path`; that a HF repo id, a local path, and `s3://` / `gs://` / `az://` are **not** claimed (the regression that matters most); `strip_oci_scheme` round-trips; the resolve contract plus eight malformed-output cases; every `LLMMAN_HOST` form incl. wildcard-to-loopback; binary default/override and the missing-binary error; empty reference rejected without touching the daemon; and the scheme stripped before hand-off with progress wired.

Verified honestly:

- [x] **23 logic assertions executed standalone against a live loopback daemon** (scheme handling incl. every non-claimed shape, delegation with the bare reference, endpoint parsing incl. wildcard rewriting, `/api/version`, pull success + progress + request body, in-band error, stream without success, resolve contract) -- all pass
- [x] `ruff format` and `ruff check` clean on all touched files
- [ ] The pytest files were **not executed in-tree**: importing `aphrodite.config.model` needs torch and the full stack, unavailable here. Flagging rather than implying coverage I do not have.
- [ ] No end-to-end run against a live `llmman serve` backed by a real registry.

> Disclosure: written with AI assistance.

### DIFF
--- a/aphrodite/config/model.py
+++ b/aphrodite/config/model.py
@@ -47,6 +47,7 @@ from aphrodite.transformers_utils.model_arch_config_convertor import (
     MODEL_ARCH_CONFIG_CONVERTORS,
     ModelArchConfigConvertorBase,
 )
+from aphrodite.transformers_utils.oci_utils import is_oci_uri, resolve_oci_model
 from aphrodite.transformers_utils.runai_utils import ObjectStorageModel, is_runai_obj_uri
 from aphrodite.transformers_utils.utils import maybe_model_redirect
 from aphrodite.utils.import_utils import LazyLoader
@@ -942,6 +943,14 @@ class ModelConfig:
         if self.model_weights:
             return
 
+        # A CNCF ModelPack artifact is pulled from a container registry and
+        # extracted to a local directory, which the default HuggingFace-format
+        # loading then sees. Handled before the object-storage path since the
+        # two schemes are disjoint.
+        if is_oci_uri(model) or is_oci_uri(tokenizer):
+            self.maybe_pull_model_tokenizer_for_oci(model, tokenizer)
+            return
+
         if not (is_runai_obj_uri(model) or is_runai_obj_uri(tokenizer)):
             return
 
@@ -974,6 +983,28 @@ class ModelConfig:
                 ignore_pattern=["*.pt", "*.safetensors", "*.bin", "*.tensors", "*.pth"],
             )
             self.tokenizer = object_storage_tokenizer.dir
+
+    def maybe_pull_model_tokenizer_for_oci(self, model: str, tokenizer: str) -> None:
+        """Pull a CNCF ModelPack artifact from an OCI registry.
+
+        The whole image is extracted to one directory, so a tokenizer naming
+        the same reference reuses it rather than pulling twice.
+
+        Args:
+            model: Model name or path
+            tokenizer: Tokenizer name or path
+        """
+        resolved: dict[str, str] = {}
+
+        if is_oci_uri(model):
+            resolved[model] = resolve_oci_model(model)
+            self.model_weights = model
+            self.model = resolved[model]
+
+        if is_oci_uri(tokenizer):
+            if tokenizer not in resolved:
+                resolved[tokenizer] = resolve_oci_model(tokenizer)
+            self.tokenizer = resolved[tokenizer]
 
     def _get_encoder_config(self) -> dict[str, Any] | None:
         return get_sentence_transformer_tokenizer_config(self.model, self.revision)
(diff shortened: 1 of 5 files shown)


aphrodite/config/model.py
```diff
@@ -942,6 +943,14 @@ def maybe_pull_model_tokenizer_for_runai(self, model: str, tokenizer: str) -> No
         if self.model_weights:
             return
 
+        # A CNCF ModelPack artifact is pulled from a container registry and
+        # extracted to a local directory, which the default HuggingFace-format
+        # loading then sees. Handled before the object-storage path since the
+        # two schemes are disjoint.
+        if is_oci_uri(model) or is_oci_uri(tokenizer):
```
AlpinDale, line 950: Maybe a bit of a nit, but this seems to break when used together with S3, for example OCI model + S3 tokenizer, and S3 model + OCI tokenizer both fail.

AlpinDale (changes requested):
Thanks for the PR! Please resolve the merge conflicts and take a look at the review comments below.

Also as an aside, I took a look through the llmman repo, and it seems to download through the daemon selected by `LLMMAN_HOST`, then runs a local CLI to find the files, which seems to [read its own local store](https://github.com/llmmanorg/llmman/blob/main/src/cmd/resolve.rs) instead of the remote daemon's...? Please let me know if that's intentional because that'd imply remote daemons (even even local ones running under another user) won't work without shared storage.

PR #1775, opened 2026-09-06
fix(exl3): give the MoE down projection its own Hadamard scratch

Andybui1012:
## Problem

`exl3_mgemm` reads `A` and writes the Hadamard-transformed input to `A_had`. The cooperative-kernel autotuner (`CoopKernelAutotuner::launch`) re-launches the kernel on the same arguments while it times candidates, so when `A_had` aliases `A`, every launch after the first transforms the previous launch's output.

`Exl3MoEMethod._apply_single_token` and `_apply_small_batch` passed `layer.exl3_small_interm_a` as both `A` and `A_had` for the down projection. Result: the first call for a given expert shape returns garbage (normalized max error ~1.0 against a dense reference), and every later call, served from the autotune cache, is correct. In serving that is the first decode step for each MoE expert shape, or for every shape whose autotune record is not already in `~/.cache/exllamav3/autotune`.

## Fix

Allocate a dedicated `exl3_small_interm_a_had` buffer next to `exl3_small_interm_a` in `process_weights_after_loading` and pass it as `A_had` at both call sites. ExLlamaV3 uses a separate buffer for this (`block_sparse_mlp.py`: "A_had must not alias A (the autotuner relaunches on the first call)"). The gate/up projections already use `exl3_small_yh` / `exl3_small_yh_gu` as separate scratch.

## Test

`tests/kernels/quantization/test_exl3_moe_small_batch.py` builds a small synthetic EXL3 MoE layer through `Exl3MoEMethod.process_weights_after_loading`, uses intermediate sizes no other test autotunes, and checks the **first** `apply()` against a per-expert fp32 dense reference (reconstructed weights with the 128-Hadamard and sign flips folded in), for both the single-token (`rows=1`) and small-batch (`rows=3`) paths. It also checks that the second call equals the first.

On the previous code: `2 failed` (normalized error 1.22 and 0.97). With the fix: `2 passed`. Verified in a CUDA 13 container on a GB10 (sm_121); the test needs a CUDA device and about 100 MiB of GPU memory.

`ruff check` and `ruff format --check` pass on both files.

### DIFF
--- a/tests/kernels/quantization/test_exl3_moe_small_batch.py
+++ b/tests/kernels/quantization/test_exl3_moe_small_batch.py
@@ -1,0 +1,141 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the Aphrodite project
+"""First-call correctness of the EXL3 MoE single-token and small-batch paths.
+
+``exl3_mgemm`` reads ``A`` and writes the Hadamard-transformed copy to
+``A_had``. The cooperative-kernel autotuner re-launches the kernel on the same
+arguments while it times candidates, so if ``A_had`` aliases ``A`` every launch
+after the first transforms the previous launch's output: the first call for a
+shape returns garbage, and later calls (served from the autotune cache) are
+correct. The down projection used to pass ``exl3_small_interm_a`` for both.
+"""
+
+import math
+from types import SimpleNamespace
+
+import pytest
+import torch
+
+from aphrodite import _custom_ops as ops
+from aphrodite.platforms import current_platform
+
+if not current_platform.is_cuda():
+    pytest.skip(reason="EXL3 requires CUDA", allow_module_level=True)
+
+DEVICE = "cuda"
+K_BITS = 3
+
+
+def _trellis(size_k: int, size_n: int) -> torch.Tensor:
+    return torch.randint(-(2**15), 2**15, (size_k // 16, size_n // 16, 16 * K_BITS), dtype=torch.int16, device=DEVICE)
+
+
+def _signs(size: int) -> torch.Tensor:
+    return torch.where(torch.rand(size, device=DEVICE) < 0.5, -1.0, 1.0).to(torch.float16)
+
+
+def _sylvester_128() -> torch.Tensor:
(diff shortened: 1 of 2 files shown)


tests/kernels/quantization/test_exl3_moe_small_batch.py
```diff
@@ -0,0 +1,141 @@
...
+            w2 = _dense_weight(
+                *(getattr(layer, f"w2_{a}").exl3_tensors[(expert_id, "w2")] for a in ("trellis", "suh", "svh"))
+            )
+            act = (torch.nn.functional.silu(x[row].float() @ w1) * (x[row].float() @ w3)).half().float()
+            out[row] += float(topk_weights[row, slot]) * (act @ w2)
+    return out
+
+
+@pytest.mark.parametrize(("rows", "intermediate"), [(1, 384), (3, 640)])
+def test_moe_small_batch_first_call_matches_dense_reference(rows: int, intermediate: int, monkeypatch, tmp_path):
```
AlpinDale, line 117:
```suggestion
def test_moe_first_call_matches_dense_reference(rows: int, intermediate: int, monkeypatch, tmp_path):
```

Still not happy with it, but this name is a lot better.

tests/kernels/quantization/test_exl3_moe_small_batch.py
```diff
@@ -0,0 +1,141 @@
...
+    """rows=1 takes ``_apply_single_token``, rows=3 ``_apply_small_batch``; both use exl3_mgemm.
+
+    The first ``apply`` below must be the call that autotunes the down
+    projection, and it must already be correct. The autotune key covers the
+    weight shape, so each case uses an intermediate size no other case (or
+    test) shares. The on-disk autotune cache is read once per process; the
+    environment override only helps when this test runs first, and is set so
+    that running the file on its own never sees a pre-seeded shape.
+    """
+    monkeypatch.setenv("EXLLAMAV3_TUNE_CACHE", str(tmp_path / "coop_autotune_v1.bin"))
```
AlpinDale, line 127: This would make the test pass unconditionally with a warm cache, but I'd say it's mostly a nit.

AlpinDale (changes requested):

PR #1785, opened 2026-09-11
chore: change aphrodite ascii art to sonar

lucyknada:
few notes:
- regular utils.py logo output didnt seem to be in use anywhere (and drifted?); removed it
- changed api_utils hardcoded width to use the same dynamic width calc old utils.py did
- I did not run full aphrodite to confirm the ascii art doesn't now height shift other log lines; but outputting it with dummy data looks identical, now just taking up 1 more line of height


AlpinDale (approved):
LGTM

